PR #20, opened 2014-12-30
Fix when building with template-haskell-2.10.0.0

RyanGlScott:
`template-haskell-2.10.0.0` provides instances for `Ord Lit` and `Lift Word8`, so only define those instances if using a version of `template-haskell` older than 2.10.0.0.


goldfirere:
Thanks!

PR #22, opened 2015-03-04
Convert (~) back to EqualityT in typeToTH

ddssff:
This corresponds to

    dsType EqualityT = return $ DConT ''(~)
    #endif

in the dsType function.


goldfirere:
I've done a little refactoring, but this is a good idea. Thanks!

ddssff:
Ah, very nice.

PR #23, opened 2015-03-07
Use th-orphans instead of th-lift and custom Ord instances

ddssff:


ddssff:
Its worth noting that haskell-src-meta uses th-orphans, and instances being what they are you can't really mix th-orphans and th-lift.

ddssff:
Oh wait, th-lift is a dependency of th-orphans.  Now I'm confused.

goldfirere:
I think this is a move in the right direction, but the patch as submitted also removes a bunch of non-redundant instances: all the `Lift` instances for the `Dxxx` types.

ddssff:
No wonder it compiled!

ddssff:
Should I refactor this pull request?

ddssff:
Yeah, it needs to be redone - it botches the 7.6.3 compatibility.

PR #24, opened 2015-03-07
Add some tests of typeToTH

ddssff:
These should be fixed by pull request 23.


ddssff:
These tests are now checked into the other pull request and slightly extended.

PR #25, opened 2015-03-16
Fix for sweetening named types

ddssff:


goldfirere:
This is subsumed by #27.

PR #29, opened 2015-03-31
DsMonad instances for ReaderT, StateT, and WriterT

ddssff:
More orphans, if you'll have them kind sir...


goldfirere:
Now released at 1.5.3. Thanks!

PR #30, opened 2015-03-31
Link test suite to in-place library

ddssff:
This is what I do when I'm avoiding real work...


goldfirere:
Thanks for this. I never quite knew (or bothered enough to learn) how to do make this more efficient.

PR #35, opened 2015-11-13
Build and pass tests on GHC 8

christiaanb:
Trying to get my own libraries to build against GHC 8, I needed to update dependencies also. `th-desugar` took the longest to port, so I thought I'd share my results.

See this PR more as a guideline/head start on actually getting `th-desugar` to build on GHC 8, not as something that needs to be merged.

### DIFF
--- a/Language/Haskell/TH/Desugar.hs
+++ b/Language/Haskell/TH/Desugar.hs
@@ -28,6 +28,9 @@ module Language.Haskell.TH.Desugar (
   DCon(..), DConFields(..), DStrictType, DVarStrictType, DForeign(..),
   DPragma(..), DRuleBndr(..), DTySynEqn(..), DInfo(..), DInstanceDec,
   Role(..), AnnTarget(..),
+#if __GLASGOW_HASKELL__ > 710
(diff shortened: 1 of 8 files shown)


goldfirere:
Thanks for doing this! If this passes the tests, I will merge but not release, as I want to take a more thorough pass through your edits than I have time for. In the meantime, others (myself included, actually) want th-desugar to build on HEAD, and this gets that for them.

goldfirere:
I'm surprised this didn't build on Travis against HEAD. Isn't that the point? Or do I see from the error messages that the testsuite dependencies didn't build?

christiaanb:
It's not building on head, because dependencies (of the test suite), sush as the async package, cannot be build on head in their current state. They will also need updating.

christiaanb:
Most of the (transitive) dependencies need trivial patches. The reason I posted this pull request specifically is because this patch is sort of non-trivial and thought it would be handy for people not having to repeat my work. I'm sorry if I let ppl believe that with this patch alone one would be able to compile and test th-desugar on head.

goldfirere:
Thanks!

PR #36, opened 2015-11-16
Fix free type variable calculation for type wildcards

christiaanb:
In my _build-on-ghc8_ PR, I for some reason thought that named wild cards in types should be treated like Tycons. Of course, named wild cards are just (free) variables, and this patch fixes that.

### DIFF
--- a/Language/Haskell/TH/Desugar.hs
+++ b/Language/Haskell/TH/Desugar.hs
@@ -180,7 +180,7 @@ fvDType = go
     go (DConT _)               = S.empty
     go DArrowT                 = S.empty
     go (DLitT {})              = S.empty
-    go (DWildCardT _)          = S.empty -- FIXME: Wildcards cannot be bound, but are they free?
+    go (DWildCardT n_m)        = maybe S.empty S.singleton n_m
 
 dtvbName :: DTyVarBndr -> S.Set Name
 dtvbName (DPlainTV n)    = S.singleton n
@@ -194,7 +194,7 @@ fvDKind = go
     go (DConK _ kis)       = foldMap fvDKind kis
     go (DArrowK k1 k2)     = go k1 `S.union` go k2
     go DStarK              = S.empty
-    go (DWildCardK _)      = S.empty -- FIXME: Wildcards cannot be bound, but are they free? 
+    go (DWildCardK n_m)    = maybe S.empty S.singleton n_m
 
 -- | Produces 'DLetDec's representing the record selector functions from
 -- the provided 'DCon'.


goldfirere:
If you're convinced, I'm happy to merge this for now. But I worry that wildcards are a bit different from ordinary free variables. In particular, I'm worried about unnamed wildcards. If we want to return wildcards as free, then we really should name (uniquely, somehow) and return all unnamed wildcards.

christiaanb:
My reasoning was: named wildcards can be bound, so that must mean they can also be free.

Though I must agree, named wild cards are weird, see for example the following `ghci` session:

```
/local/home/baaijcpr/singletons (ghc8)$ ghci -XTemplateHaskell -XNamedWildCards -XScopedTypeVariables
GHCi, version 7.11.20151111: http://www.haskell.org/ghc/  :? for help
Prelude> :m +Language.Haskell.TH
Prelude Language.Haskell.TH> runQ [d|foo :: a -> a;foo x = x|]
[SigD foo_0 (ForallT [PlainTV a_1] [] (AppT (AppT ArrowT (VarT a_1)) (VarT a_1))),FunD foo_0 [Clause [VarP x_2] (NormalB (VarE x_2)) []]]
Prelude Language.Haskell.TH> runQ [d|foo :: _a -> _a;foo x = x|]
[SigD foo_3 (AppT (AppT ArrowT (WildCardT (Just _a_1627405429))) (WildCardT (Just _a_1627405429))),FunD foo_3 [Clause [VarP x_4] (NormalB (VarE x_4)) []]]
Prelude Language.Haskell.TH> runQ [d|foo :: forall _a . _a -> _a;foo x = x|]
[SigD foo_6 (ForallT [PlainTV _a_5] [] (AppT (AppT ArrowT (WildCardT (Just _a_5))) (WildCardT (Just _a_5)))),FunD foo_6 [Clause [VarP x_7] (NormalB (VarE x_7)) []]]
Prelude Language.Haskell.TH> runQ [d|foo :: forall _ . _ -> _;foo x = x|]

<interactive>:6:23: error: parse error on input \u2018_\u2019
```

So it seems GHC by default, put a `forall` for _normal_ type variables (decl 1), but doesn't put a `forall` by default when we're dealing with named wild cards (decl 2). However, named wild cards can be bound (decl 3). We can however not bind unnamed wild cards (decl 4).

So because unnamed wild cards cannot be bound, they cannot be free either.

goldfirere:
I think your (decl 3) is being mishandled by TH and is a GHC bug. That declaration is just a normal forall, binding a type variable whose name happens to start with an underscore. Note that a plain `_` is not considered a lower-case identifier in Haskell, but essentially a keyword. So I'm not surprised the last one fails. TH should not use `WildCardT` in case 3.

goldfirere:
Reported as GHC's #11098

christiaanb:
If decl 3 is a bug, then I guess named wildcards cannot be bound, meaning they probably cannot be free either. I based this patch on the assumption that the given ghci session was correct. If it is not, then this PR should be reverted until we know exactly whether named wild cards can be considered as free variables or not.

PR #38, opened 2016-01-19
Build on GHC 8.0-rc1

christiaanb:
Currently has the following test-suite failures on GHC 8.0.1-rc1:

      Test/Run.hs:275:
      1) th-desugar library reifies local definition 2

      Test/Run.hs:275:
      2) th-desugar library reifies local definition 3

Basically, this is due to how GHC is now treating implicitly bound type variables differently it seems. Given:

```haskell
  class R2 a b where
    r3 :: a -> b -> c -> a
    type R4 b a :: *
    data R5 a :: *
```

reifying `r3`  in GHC-7.10.3 will give something along the lines of:

```haskell
r3 :: forall a b . R2 a b => forall c . a -> b -> c -> a
```

and `th-desugar` would give the same info.

However, in GHC 8-rc1, reifying `r3` gives something along the lines of:

```haskell
r3 :: forall a b . R2 a b => a -> b -> c -> a
```

where `th-desugar` still gives:

```haskell
r3 :: forall a b . R2 a b => forall c . a -> b -> c -> a
```

That is the testsuite failure you are seeing. If you have any hints where to look for a solution, I'm happy to investigate. I'm sending this pull request nonetheless just to let you know that I've already done some work on porting `th-desugar` over to GHC 8.0-rc1.

The branch can be build and tested with the following `stack.yaml`: https://gist.github.com/christiaanb/2c94f162c706fca7ee76

Another weird thing is that I needed to annotate some TH quotes with their types, for example, the original:

```haskell
test3_lamcase = [| foldr (-) 0 (map (\case { Just x -> x ; Nothing -> (-3) }) [Just 1, Nothing, Just 19, Nothing]) |]
```

gives the following error:

```
   Test/Run.hs:63:53: error:
       • Couldn't match type ‘Maybe Integer’ with ‘forall a. Maybe a’
         Expected type: [forall a. Maybe a]
           Actual type: [Maybe Integer]
       • In the second argument of ‘(:)’, namely
           ‘(:) (Just 19) ((:) (Nothing :: Maybe Integer) [])’
         In the second argument of ‘(:)’, namely
           ‘(:) Nothing ((:) (Just 19) ((:) (Nothing :: Maybe Integer) []))’
         In the second argument of ‘map’, namely
           ‘(:)
              (Just 1)
              ((:) Nothing ((:) (Just 19) ((:) (Nothing :: Maybe Integer) [])))’
```

I had to annotate it like so:

```haskell
test3_lamcase = [| foldr (-) 0 (map (\case { Just x -> x ; Nothing -> (-3) }) [Just 1, Nothing :: Maybe Integer, Just 19, Nothing :: Maybe Integer]) |]
```

in order for it to succesfully compile. I haven't been able to reduce it to a small test-case yet.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,14 @@
 `th-desugar` release notes
 ==========================
 
+Version 1.6
+-----------
+* Work with GHC 8, with thanks to @christiaanb for getting this change going.
+
+* `DKind` is merged with `DType`.
+
+* `Generic` instances for everything.
+
 Version 1.5.5
 -------------
 
--- a/Language/Haskell/TH/Desugar.hs
+++ b/Language/Haskell/TH/Desugar.hs
@@ -23,7 +23,7 @@ eir@cis.upenn.edu
 
 module Language.Haskell.TH.Desugar (
   -- * Desugared data types
-  DExp(..), DLetDec(..), DPat(..), DType(..), DKind(..), DCxt, DPred(..),
+  DExp(..), DLetDec(..), DPat(..), DType(..), DKind, DCxt, DPred(..),
   DTyVarBndr(..), DMatch(..), DClause(..), DDec(..), NewOrData(..),
   DTypeFamilyHead(..), DFamilyResultSig(..), InjectivityAnn(..),
   DCon(..), DConFields(..), DBangType, DVarBangType,
@@ -113,10 +113,6 @@ instance Desugar Exp DExp where
 instance Desugar Type DType where
   desugar = dsType
   sweeten = typeToTH
-
-instance Desugar Kind DKind where
-  desugar = dsKind
-  sweeten = kindToTH
 
 instance Desugar Cxt DCxt where
   desugar = dsCxt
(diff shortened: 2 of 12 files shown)


christiaanb:
The weird error is probably: https://ghc.haskell.org/trac/ghc/ticket/11458

goldfirere:
I took at look at this today. Many thanks for your contribution! But the way you've adapted th-desugar to GHC 8 violates some of the philosophy behind the package. My goal is to keep the core data structures identical across GHC versions. In this way, a client of TH that uses th-desugar need not be littered with tons of CPP to get it all to work together. Unfortunately, this means more heavy lifting for th-desugar. I've started to put this in. But I've run out of time today, I'm afraid. I stopped at `RAE was here` in Core.hs. I plan on picking this up again, quite likely before the final release of GHC 8. In the meantime, I'm happy to collaborate on getting this all working.

The next step is to build a quick-and-dirty type unifier to be able to convert GADT syntax to regular syntax.

Wait! No! Don't do that! Instead, drop Haskell98 syntax from th-desugar and just use GADT syntax. It's really really easy to convert from Haskell98 to GADT, but the other way is very very painful. So let's just do the easy thing. I don't think it would be hard for downstream clients to adapt. But I'm still out of time today. :(

christiaanb:
Ah, yes, sorry, I forgot to say that this pull request was not meant as "please merge", but more like "this is the quick hack to get stuff to compile on ghc8-rc1".

With that said, I'm willing to make a patch that is completely proper, one that adheres to the philosophy of the package. I can do this completely myself, with you just providing comments when I've written new code, or we work together on the same branch. I'm cool either way.

One question about a patch that adheres to the philosophy of the package: the th-desugar `Core` data types share commonalities with TH version 2.10. Now that we're "upgrading" to TH version 2.11, is the idea to follow some of the changes in 2.11 and be "backward" compatible with 2.10? Or, do we keep the 2.10-like data types, and be "forward" compatible with 2.11?

Also, now that you've mention GADTs, it seems this is valid GADT syntax:

```haskell
data T where
   MkT1, MkT2 :: T
```

which TH-2.11 supports like:

```haskell
GadtC [Name] [BangType] Type
```

consequently, I had to make `dsCon` change from: `dsCon :: Con -> DCon`, to: `dsCon :: Con -> [DCon]. If we are adopting the GADT syntax in `th-desugar`, I assume we do _not_ keep this weird multiple names aspect of TH 2.11.

christiaanb:
Unlike what my patches currently reflect, I think I would be in favour of being "forward" compatible with TH-2.11 (with the exception of the GADT syntax); i.e. try to keep the th-desugar `Core` data types as much the same as they are now.

goldfirere:
Thanks, Christiaan, for volunteering to help out here. Very much appreciated.

Here are some of my philosophical thoughts:
* `th-desugar` should remain compatible with GHCs back to 7.6.
* If `th-desugar` is to cope with new TH features, its AST will have to change. I'm not terribly concerned about AST change between `th-desugar` versions, though fixing a `th-desugar` version and varying the GHC version should *not* cause a change in AST.
* I have a tiny preference for mirroring the latest TH AST as opposed to accumulating historical decisions that make for a sub-optimal `th-desugar` AST.
* Design choices should be made thinking what would be simplest for clients to consume. Let `th-desugar` do the heavy lifting so that clients don't have to.

My recent push to the `ghc8` branch on my repo embodies some of these decisions. For example, I favor `DTypeFamilyHead`, echoing the recent TH. But I wholeheartedly agree that one `DCon` should mean one constructor, unlike the recent TH.

Also, you'll see that I include data structures from the recent TH in `th-desugar`, like `InjectivityAnn`. This is to keep consistency across GHC versions.

Does this give enough direction?

christiaanb:
Yes, this should provide enough guidance towards making a proper patch. I'm fixing a bug in my own software right now, but I'll start working on th-desugar afterwards (most likely tomorrow).

PR #44, opened 2016-10-01
Adapt to template-haskell changes in GHC HEAD deriving strategies.

mboes:
This patch does not add support for them in th-desugar. But it doesn't
make the situation any worse either. At least with this patch
th-desugar now compiles, and TH code that doesn't mention strategies
gets desugared the same as before.

cc @RyanGIScott

### DIFF
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -9,6 +9,7 @@ processing. The desugared types and constructors are prefixed with a D.
 
 {-# LANGUAGE TemplateHaskell, LambdaCase, CPP, DeriveDataTypeable,
              DeriveGeneric, TupleSections #-}
+{-# OPTIONS_GHC -fno-warn-dodgy-imports #-}
 
 module Language.Haskell.TH.Desugar.Core where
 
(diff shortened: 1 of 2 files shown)


goldfirere:
I'm not all that excited to merge this PR, given that it would not be hard to add proper support for deriving strategies in th-desugar. And this patch does make things worse in that it will introduce warnings for older GHC versions.

Would you care to do the full implementation? Thanks!

mboes:
> Would you care to do the full implementation? Thanks!

Not in the short term, I'm afraid. Already exceeded my time budget for the weekend working on this. Full support for strategies would be great, but I didn't find it so easy to deal with yet another template-haskell breaking API change rippling through th-desugar in a clean way (yet more nested CPP awaits...). IMHO at least compiling is by definition infinitely better than failing to build altogether. So I'm not sure I see an issue with merging this now and then merging something better later. :) Especially given that this PR induces no loss of functionality and strategies in TH currently have no users (clearly not any current th-desugar users at any rate, since they previously didn't exist!).

If the issue is just warnings, that's much easier to fix than a full-blown support for strategies. And those warnings would appear even with full-blown support for strategies anyways (it's an orthogonal issue, due to new conflicting names with template-haskell, so need to hide more when importing).

Warnings fixed in the latest commit. I opted for turning off -fwarn-dodgy-imports rather than using yet more CPP to fiddle with the hiding list. See the commit message for the rationale.

goldfirere:
OK. Fair enough. Thanks much for this contribution!

Issue #43, opened 2016-09-30
expand incorrectly expands closed type family 


RyanGlScott:
In the code below, `th-desugar` thinks `F Bool` reduces to `Int`, when it actually reduces to `Char`:

``` haskell
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeFamilies #-}
module THDesugarBug where

import Language.Haskell.TH
import Language.Haskell.TH.Desugar

type family F (a :: k) :: * where
    F (a :: * -> *) = Int
    F (a :: k)      = Char

-- Proof that this should reduce to a Char
c :: F Bool
c = 'c'

$(return [])

main :: IO ()
main = putStrLn $([t| F Bool |] >>= dsType >>= expand >>= stringE . pprint . typeToTH)
```

```
$ /opt/ghc/8.0.1/bin/runghc THDesugarBug.hs 
GHC.Types.Int
```

My understanding of the documentation of `expand` and `expandUnsoundly` is that this kind of thing should only happen with the latter function. That is, `expand` shouldn't expand the above example at all, since it distinguishes cases by kind.

goldfirere:
This is actually GHC's fault. I've filed GHC#12646.

RyanGlScott:
[GHC Trac #12646](https://ghc.haskell.org/trac/ghc/ticket/12646) will be fixed in GHC 8.2. So as long as you are willing to use a recent enough GHC HEAD, the code above will work properly:

```
λ> :set -XPolyKinds -XTemplateHaskell -XTypeFamilies 
λ> import Language.Haskell.TH
λ> import Language.Haskell.TH.Desugar
λ> type family F (a :: k) :: * where F (a :: * -> *) = Int; F (a :: k) = Char
λ> putStrLn $([t| F Bool |] >>= dsType >>= expand >>= stringE . pprint . typeToTH)
Ghci1.F GHC.Types.Bool
```

PR #46, opened 2016-12-06
Implement deriving strategies in th-desugar

### LINKED ISSUE #45: Support deriving strategies
See also #44.

RyanGlScott:
Fixes #45.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,10 @@ env:
  - CABALVER=1.22 GHCVER=7.10.2
  - CABALVER=1.22 GHCVER=7.10.3
  - CABALVER=1.24 GHCVER=8.0.1
-# HEAD has TH updates - CABALVER=1.24 GHCVER=head   # see section about GHC HEAD snapshots
+ - CABALVER=head GHCVER=head   # see section about GHC HEAD snapshots
+
+allow_failures:
+ - env: CABALVER=head GHCVER=head
 
 # Note: the distinction between `before_install` and `install` is not important.
 before_install:
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,9 @@ env:
  - CABALVER=1.24 GHCVER=8.0.1
  - CABALVER=head GHCVER=head   # see section about GHC HEAD snapshots
 
-allow_failures:
- - env: CABALVER=head GHCVER=head
+matrix:
+  allow_failures:
+   - env: CABALVER=head GHCVER=head
 
 # Note: the distinction between `before_install` and `install` is not important.
 before_install:
--- a/Language/Haskell/TH/Desugar.hs
+++ b/Language/Haskell/TH/Desugar.hs
@@ -51,6 +51,9 @@ module Language.Haskell.TH.Desugar (
   dsTypeFamilyHead, dsFamilyResultSig,
 #endif
   dsDerivClause,
+#if __GLASGOW_HASKELL__ >= 801
+  dsDerivStrategy,
+#endif
 
   -- * Converting desugared AST back to TH AST
   module Language.Haskell.TH.Desugar.Sweeten,
--- a/Language/Haskell/TH/Desugar.hs
+++ b/Language/Haskell/TH/Desugar.hs
@@ -25,7 +25,7 @@ module Language.Haskell.TH.Desugar (
   -- * Desugared data types
   DExp(..), DLetDec(..), DPat(..), DType(..), DKind, DCxt, DPred(..),
   DTyVarBndr(..), DMatch(..), DClause(..), DDec(..),
-  DDerivClause(..), DDerivStrategy(..),
+  DDerivClause(..), DerivStrategy(..),
   Overlap(..), NewOrData(..),
   DTypeFamilyHead(..), DFamilyResultSig(..), InjectivityAnn(..),
   DCon(..), DConFields(..), DBangType, DVarBangType,
@@ -44,15 +44,11 @@ module Language.Haskell.TH.Desugar (
   dsCon, dsForeign, dsPragma, dsRuleBndr,
 
   -- ** Secondary desugaring functions
-  PatM, dsPred, dsPat, dsDec, dsLetDec,
+  PatM, dsPred, dsPat, dsDec, dsDerivClause, dsLetDec,
   dsMatches, dsBody, dsGuards, dsDoStmts, dsComp, dsClauses,
   dsBangType, dsVarBangType,
 #if __GLASGOW_HASKELL__ > 710
   dsTypeFamilyHead, dsFamilyResultSig,
-#endif
-  dsDerivClause,
-#if __GLASGOW_HASKELL__ >= 801
-  dsDerivStrategy,
 #endif
 
   -- * Converting desugared AST back to TH AST
--- a/Language/Haskell/TH/Desugar.hs
+++ b/Language/Haskell/TH/Desugar.hs
@@ -25,6 +25,7 @@ module Language.Haskell.TH.Desugar (
   -- * Desugared data types
   DExp(..), DLetDec(..), DPat(..), DType(..), DKind, DCxt, DPred(..),
   DTyVarBndr(..), DMatch(..), DClause(..), DDec(..),
+  DDerivClause(..), DDerivStrategy(..),
   Overlap(..), NewOrData(..),
   DTypeFamilyHead(..), DFamilyResultSig(..), InjectivityAnn(..),
   DCon(..), DConFields(..), DBangType, DVarBangType,
@@ -49,6 +50,7 @@ module Language.Haskell.TH.Desugar (
 #if __GLASGOW_HASKELL__ > 710
   dsTypeFamilyHead, dsFamilyResultSig,
 #endif
+  dsDerivClause,
 
   -- * Converting desugared AST back to TH AST
   module Language.Haskell.TH.Desugar.Sweeten,
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -9,17 +9,13 @@ processing. The desugared types and constructors are prefixed with a D.
 
 {-# LANGUAGE TemplateHaskell, LambdaCase, CPP, DeriveDataTypeable,
              DeriveGeneric, TupleSections #-}
-{-# OPTIONS_GHC -fno-warn-dodgy-imports #-}
 
 module Language.Haskell.TH.Desugar.Core where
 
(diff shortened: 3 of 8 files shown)


Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -233,9 +236,18 @@ data DInfo = DTyConI DDec (Maybe [DInstanceDec])
...
-data DDerivClause = DDerivClause (Maybe DerivStrategy) DCxt
-#endif
+-- | Corresponds to TH's @DerivClause@ type.
+data DDerivClause = DDerivClause (Maybe DDerivStrategy) DCxt
+                  deriving (Show, Typeable, Data, Generic)
+
+-- | Corresponds to TH's @DerivStrategy@ type.
+
+-- This is redefined here, since the Newtype constructor of DerivStrategy
+-- conflicts with the Newtype constructor of NewOrData.
```
RyanGlScott: I'm starting to regret naming that thing `Newtype` now... :)

.travis.yml
```diff
@@ -5,7 +5,7 @@ sudo: required
 
 # The following enables several GHC versions to be tested; often it's enough to test only against the last release in a major GHC version. Feel free to omit lines listings versions you don't need/want testing for.
 env:
- - CABALVER=1.18 GHCVER=7.6.3
+ - CABALVER=1.24 GHCVER=7.6.3
```
goldfirere: Why this change? Seems odd. Compatibility over a range of GHCs/cabals is important to th-desugar.

Language/Haskell/TH/Desugar.hs
```diff
@@ -25,6 +25,7 @@ module Language.Haskell.TH.Desugar (
   -- * Desugared data types
   DExp(..), DLetDec(..), DPat(..), DType(..), DKind, DCxt, DPred(..),
   DTyVarBndr(..), DMatch(..), DClause(..), DDec(..),
+  DDerivClause(..), DDerivStrategy(..),
```
goldfirere: We don't need `DDerivStrategy`, as we can just use TH's `DerivStrategy`.

Never mind. I see why you did this. Argh. But I don't have a better solution.

Haskell needs namespaces!

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -797,14 +810,17 @@ dsDec (ClosedTypeFamilyD n tvbs m_k eqns) = do
 dsDec (RoleAnnotD n roles) = return [DRoleAnnotD n roles]
 #endif
 #if __GLASGOW_HASKELL__ >= 709
-#if __GLASGOW_HASKELL__ >= 801
-dsDec (StandaloneDerivD (Just _) cxt ty) = fail "Deriving strategies not supported by th-desugar."
-dsDec (StandaloneDerivD Nothing cxt ty) =
-#else
+# if MIN_VERSION_template_haskell(2,12,0)
```
goldfirere: Please be consistent with local style and use `#if`, without the space. Thanks!

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -15,6 +15,9 @@ module Language.Haskell.TH.Desugar.Core where
 
 import Prelude hiding (mapM, foldl, foldr, all, elem, exp, concatMap, and)
 
+#if MIN_VERSION_template_haskell(2,12,0)
```
goldfirere: Please use `#if __GLASGOW_HASKELL__ >= 801` (here and elsewhere). That way, I can compile without cabal.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1043,13 +1059,23 @@ dsTvb (KindedTV n k) = DKindedTV n <$> dsType k
...
-dsDerivClause (DerivClause Nothing cxt) = dsCxt cxt
-dsDerivClause (DerivClause _ cxt) = fail "Deriving strategies not supported in th-desugar."
+#if MIN_VERSION_template_haskell(2,12,0)
+-- | Desugar a @DerivClause@.
+dsDerivClause :: DsMonad q => DerivClause -> q DDerivClause
+dsDerivClause (DerivClause mds cxt) =
+  DDerivClause (fmap dsDerivStrategy mds) <$> dsCxt cxt
+
+-- | Desugar a @DerivStrategy@.
+dsDerivStrategy :: DerivStrategy -> DDerivStrategy
```
goldfirere: Should this function be exported from `Language.Haskell.TH.Desugar`? I suppose so.

Test/Run.hs
```diff
@@ -201,6 +201,13 @@ test_standalone_deriving = (MkBlarggie 5 'x') == (MkBlarggie 5 'x')
 test_standalone_deriving = True
 #endif
 
+test_deriving_strategies :: Bool
```
goldfirere: Thanks for adding this test!

Language/Haskell/TH/Desugar.hs
RyanGlScott: I have considered going back into GHC and changing `Newtype` to `NewtypeStrategy`, `Stock` to `StockStrategy`, and `Anyclass` to `AnyclassStrategy`. That would be one way around this :)

.travis.yml
goldfirere: OK. Not worth the energy to understand better. Thanks.

Language/Haskell/TH/Desugar.hs
goldfirere: I'd actually be in favor of this resolution. Deriving strategies are obscure (but necessary -- not meant to diminish your work in implementing them!), and `Newtype` seems to take up a prime spot in the namespace.

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: Good idea. I also added a note to the Haddocks explaining that it's 8.2+ only.

Language/Haskell/TH/Desugar.hs
RyanGlScott: Cool, I can prepare a patch that does that.

I suppose we could put this PR on hiatus until that gets merged, or we could adopt this change now and change to the new names later. Up to you.

RyanGlScott:
@goldfirere: Now that http://git.haskell.org/ghc.git/commit/5349d648fd7af3f50953e8594b3d148ab073017f is in GHC HEAD, I've updated this PR so that we reuse the `DerivStrategy` datatype from TH. Much cleaner!

One small problem is that the testsuite doesn't compile on GHC HEAD:

```
[4 of 4] Compiling Run              ( Test/Run.hs, dist/build/spec/spec-tmp/Run.o )

Test/Run.hs:134:14: error:
    • Couldn't match expected type ‘PolyTF IO’ with actual type ‘Bool’
    • In the pattern: True
      In an equation for ‘f_aMl3’: f_aMl3 True = ()
      In the expression:
        (let
           f_aMl3 :: PolyTF IO -> ()
           f_aMl3 True = ()
         in f_aMl3)
```

And whaddya know, turns out [this is my fault](https://github.com/goldfirere/th-desugar/blob/2ced2834e3e313970b9d029f87995cf0fa86adcf/Test/Run.hs#L134-L136) too :)

```haskell
test_e8a = $(test_expand8 >>= dsExp >>= expand >>= return . expToTH)
  -- the line above should fail once GHC#8953 is fixed for closed type
  -- families
```

What should we do with this test now that GHC Trac #8953 (and thus #43) has been fixed? Conditionally include it with CPP? Remove it? Something else?

goldfirere:
What is that test doing? Is there a way to fix it? (A faint memory says: no.) If not, then just disable it in HEAD. And, now that you've brought us up-to-date with HEAD, enable testing of HEAD with Travis, please.

Many thanks for all this work -- much appreciated!

RyanGlScott:
> What is that test doing? Is there a way to fix it? (A faint memory says: no.) If not, then just disable it in HEAD.

It seems to be testing the (faulty) type expansion of [this particular closed type family](https://github.com/goldfirere/th-desugar/blob/7e0aca246ecd63bd7c1ca75bea0332d4d15690aa/Test/Splices.hs#L247-L248):

```haskell
type family PolyTF (x :: k) :: * where
  PolyTF (x :: *) = Bool
```

Now that GHC Trac #8953 has been fixed, I'd argue that it no longer makes sense. I've disabled it `#if __GLASGOW_HASKELL__ >= 801`.

> And, now that you've brought us up-to-date with HEAD, enable testing of HEAD with Travis, please.

Done. Sadly, I encountered some more failing tests in the testsuite:

```
Failures:

  Test/Run.hs:264: 
  1) th-desugar library passes dec test 9

  Test/Run.hs:295: 
  2) th-desugar library reifies local definition 19
```

I'll have to investigate further to see what's wrong.

RyanGlScott:
OK, I see why `th-desugar library reifies local definition 19` is failing. Consider this type family from the `th-desugar` test suite:

```haskell
type family R21 (a :: k) (b :: k) :: k where R21 a b = b
```

If you reify it normally, you get kind signatures on its LHS types, as expected:

```
$ inplace/bin/ghc-stage2 --interactive
GHCi, version 8.1.20161208: http://www.haskell.org/ghc/  :? for help
λ> :set -XKindSignatures -XPolyKinds -XTypeFamilies -XTemplateHaskell
λ> import Language.Haskell.TH
λ> type family R21 (a :: k) (b :: k) :: k where R21 a b = b
λ> putStrLn $(reify ''R21 >>= stringE . pprint)
type family Ghci1.R21 (a_0 :: k_1) (b_2 :: k_1) :: k_1 where
    Ghci1.R21 (a_3 :: k_4) (b_5 :: k_4) = b_5
```

But if you quote it, you won't get the kind signatures!

```
λ> putStrLn $([d| type family R21 (a :: k) (b :: k) :: k where R21 a b = b |] >>= stringE . pprint)type family R21_0 (a_1 :: k_2) (b_3 :: k_2) :: k_2 where
    R21_0 a_4 b_5 = b_5
```

Would it be cheating to fix this test by just giving explicit kind signatures to `a` and `b` in `R21 a b = b`?

RyanGlScott:
Following your advice in https://github.com/goldfirere/th-desugar/pull/48#issuecomment-266778976, I was able to fix `th-desugar library reifies local definition 19`! And weirdly enough, I'm no longer able to trigger `th-desugar library passes dec test 9`, so I believe I've fixed all the necessary tests.

Other than the Travis build for GHC HEAD not working satisfactorily (as noted in https://github.com/goldfirere/th-desugar/pull/46#issuecomment-266575931), this PR should be ready to be merged, pending any further suggestions.

goldfirere:
Hooray!

PR #47, opened 2016-12-06
Implement type applications

RyanGlScott:
Part two of my crusade to get recent Template Haskell additions into `th-desugar`.

### DIFF
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -371,7 +371,7 @@ dsExp (StaticE exp) = DStaticE <$> dsExp exp
 #if __GLASGOW_HASKELL__ > 710
 dsExp (UnboundVarE n) = return (DVarE n)
 #endif
-#if MIN_VERSION_template_haskell(2,12,0)
+#if __GLASGOW_HASKELL__ >= 801
 dsExp (AppTypeE exp ty) = DAppTypeE <$> dsExp exp <*> dsType ty
 #endif
 
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -41,6 +41,7 @@ data DExp = DVarE Name
           | DConE Name
           | DLitE Lit
           | DAppE DExp DExp
+          | DAppTypeE DExp DType
           | DLamE [Name] DExp
           | DCaseE DExp [DMatch]
           | DLetE [DLetDec] DExp
@@ -369,6 +370,9 @@ dsExp (StaticE exp) = DStaticE <$> dsExp exp
 #endif
 #if __GLASGOW_HASKELL__ > 710
 dsExp (UnboundVarE n) = return (DVarE n)
+#endif
+#if MIN_VERSION_template_haskell(2,12,0)
+dsExp (AppTypeE exp ty) = DAppTypeE <$> dsExp exp <*> dsType ty
 #endif
 
 -- | Desugar a lambda expression, where the body has already been desugared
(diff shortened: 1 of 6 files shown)


Language/Haskell/TH/Desugar/Sweeten.hs
```diff
@@ -56,6 +56,11 @@ expToTH (DStaticE _)         = error "Static expressions supported only in GHC 7
 #else
 expToTH (DStaticE exp)       = StaticE (expToTH exp)
 #endif
+#if MIN_VERSION_template_haskell(2,12,0)
+expToTH (DAppTypeE exp ty)   = AppTypeE (expToTH exp) (typeToTH ty)
+#else
+expToTH (DAppTypeE {})       = error "Type applications supported only in GHC 8.2+"
```
RyanGlScott: An alternative design would be to make `DAppTypeE` drop the type application and just turn into a plain `AppE` on GHC < 8.2. Which do you prefer?

RyanGlScott:
Hm... now that I looked closer, I might have missed some more pattern-matches on `DType` values (for instance, in [`expand_type`](https://github.com/goldfirere/th-desugar/blob/2b848e744ca137184c2c6843faae342c24c7591b/Language/Haskell/TH/Desugar/Expand.hs#L63) and [`build_subst`](https://github.com/goldfirere/th-desugar/blob/2b848e744ca137184c2c6843faae342c24c7591b/Language/Haskell/TH/Desugar/Expand.hs#L169), and others)...

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -370,6 +371,9 @@ dsExp (StaticE exp) = DStaticE <$> dsExp exp
 #if __GLASGOW_HASKELL__ > 710
 dsExp (UnboundVarE n) = return (DVarE n)
 #endif
+#if MIN_VERSION_template_haskell(2,12,0)
```
goldfirere: As in the other review, please use the `__GLASGOW_HASKELL__` macro instead.

Language/Haskell/TH/Desugar/Sweeten.hs
goldfirere: I think the most consistent choice would be to drop the type application but otherwise succeed. There's not a well-specified behavior for this kind of thing in th-desugar, so I'm not very bothered one way or another.

Test/Splices.hs
```diff
@@ -211,6 +216,12 @@ test40_wildcards = [| let f :: (Show a, _) => a -> a -> _
                       f True False :: String |]
 #endif
 
+#if MIN_VERSION_template_haskell(2,12,0)
+test41_typeapps = [| let f :: forall a. (a -> Bool) -> Bool
+                         f g = g (undefined @(_) @(a)) in
```
goldfirere: Why the parentheses? It should be `@_ @a`, no?

RyanGlScott: The parentheses do indeed seem redundant. I mostly copied [this test](http://git.haskell.org/ghc.git/blob/c5fbbac9fdf86a94b636849e4aa6f1c4ecc9acf2:/testsuite/tests/th/T12530.hs#l14) wholesale from the GHC testsuite, which had this style (that's no excuse though, considering I probably wrote that test :) ) I'll change it.

Language/Haskell/TH/Desugar/Sweeten.hs
RyanGlScott: Very well, I've adopted the convention of dropping the applied type on old versions of `template-haskell`.

RyanGlScott:
Oops, ignore my earlier ramblings about `expand_type` and `build_subst`. I forgot that VTA doesn't require any changes to the type syntax, so nothing about type expansion needs to be changed.

I did catch a place in `Language.Haskell.TH.Desugar.Match` where a `DExp` was being pattern-matched on but `DAppTypeE` wasn't considered, so I added a case for it.

goldfirere:
Hooray! Thanks!

PR #48, opened 2016-12-10
Add support for type signatures in patterns

RyanGlScott:
In http://git.haskell.org/ghc.git/commit/b0121209f8fb47a7cb8fc32e10d8e2c06d4502c2, Template Haskell gained the ability to support type signatures in patterns. We should do the same in `th-desugar`.

### DIFF
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -56,6 +56,7 @@ data DPat = DLitPa Lit
           | DConPa Name [DPat]
           | DTildePa DPat
           | DBangPa DPat
+          | DSigPa DPat DType
           | DWildPa
           deriving (Show, Typeable, Data, Generic)
 
@@ -590,10 +591,7 @@ dsPat (ListP pats) = go pats
           h' <- dsPat h
           t' <- go t
           return $ DConPa '(:) [h', t']
-dsPat (SigP _ _) =
-  lift $ impossible
-             ("At last check (Aug 2013), type patterns in signatures are not\n" ++
-              "supported in GHC. They are not supported in th-desugar either.")
+dsPat (SigP pat ty) = DSigPa <$> dsPat pat <*> dsType ty
 dsPat (ViewP _ _) =
   fail "View patterns are not supported in th-desugar. Use pattern guards instead."
 
@@ -604,6 +602,7 @@ dPatToDExp (DVarPa name) = DVarE name
 dPatToDExp (DConPa name pats) = foldl DAppE (DConE name) (map dPatToDExp pats)
 dPatToDExp (DTildePa pat) = dPatToDExp pat
 dPatToDExp (DBangPa pat) = dPatToDExp pat
+dPatToDExp (DSigPa pat ty) = DSigE (dPatToDExp pat) ty
 dPatToDExp DWildPa = error "Internal error in th-desugar: wildcard in rhs of as-pattern"
 
 -- | Remove all wildcards from a pattern, replacing any wildcard with a fresh
@@ -614,6 +613,7 @@ removeWilds p@(DVarPa _) = return p
 removeWilds (DConPa con_name pats) = DConPa con_name <$> mapM removeWilds pats
 removeWilds (DTildePa pat) = DTildePa <$> removeWilds pat
 removeWilds (DBangPa pat) = DBangPa <$> removeWilds pat
+removeWilds (DSigPa pat ty) = DSigPa <$> removeWilds pat <*> pure ty
 removeWilds DWildPa = DVarPa <$> newUniqueName "wild"
 
 extractBoundNamesDPat :: DPat -> S.Set Name
@@ -1174,9 +1174,10 @@ isUniversalPattern (DConPa con_name pats) = do
   if length cons == 1
   then fmap and $ mapM isUniversalPattern pats
   else return False
-isUniversalPattern (DTildePa {}) = return True
-isUniversalPattern (DBangPa pat) = isUniversalPattern pat
-isUniversalPattern DWildPa       = return True
+isUniversalPattern (DTildePa {})  = return True
+isUniversalPattern (DBangPa pat)  = isUniversalPattern pat
+isUniversalPattern (DSigPa pat _) = isUniversalPattern pat
+isUniversalPattern DWildPa        = return True
 
 -- | Apply one 'DExp' to a list of arguments
 applyDExp :: DExp -> [DExp] -> DExp
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -622,6 +622,7 @@ extractBoundNamesDPat (DVarPa n)      = S.singleton n
 extractBoundNamesDPat (DConPa _ pats) = S.unions (map extractBoundNamesDPat pats)
 extractBoundNamesDPat (DTildePa p)    = extractBoundNamesDPat p
 extractBoundNamesDPat (DBangPa p)     = extractBoundNamesDPat p
+extractBoundNamesDPat (DSigPa p _)    = extractBoundNamesDPat p
 extractBoundNamesDPat DWildPa         = S.empty
 
 -- | Desugar @Info@
--- a/Language/Haskell/TH/Desugar/Match.hs
+++ b/Language/Haskell/TH/Desugar/Match.hs
@@ -142,7 +142,9 @@ tidy1 v (DBangPa pat) =
     DConPa _ _ -> tidy1 v pat   -- already strict
     DTildePa p -> tidy1 v (DBangPa p) -- discard ~ under !
     DBangPa p  -> tidy1 v (DBangPa p) -- discard ! under !
+    DSigPa p _ -> tidy1 v (DBangPa p) -- discard sig under !
     DWildPa    -> return (id, DBangPa pat)  -- no change
+tidy1 v (DSigPa pat _) = tidy1 v pat
 tidy1 _ DWildPa = return (id, DWildPa)
 
 wrapBind :: Name -> Name -> DExp -> DExp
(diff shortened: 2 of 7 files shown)


Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -590,10 +591,7 @@ dsPat (ListP pats) = go pats
           h' <- dsPat h
           t' <- go t
           return $ DConPa '(:) [h', t']
-dsPat (SigP _ _) =
-  lift $ impossible
-             ("At last check (Aug 2013), type patterns in signatures are not\n" ++
-              "supported in GHC. They are not supported in th-desugar either.")
```
RyanGlScott: AFAICT, the difficulty in supporting type signatures in patterns only concerned reifying them. Therefore, I don't _think_ it would be a problem to unconditionally expose this code.

Language/Haskell/TH/Desugar/Match.hs
```diff
@@ -142,7 +142,9 @@ tidy1 v (DBangPa pat) =
     DConPa _ _ -> tidy1 v pat   -- already strict
     DTildePa p -> tidy1 v (DBangPa p) -- discard ~ under !
     DBangPa p  -> tidy1 v (DBangPa p) -- discard ! under !
+    DSigPa p _ -> tidy1 v (DBangPa p) -- discard sig under !
```
goldfirere: Why? This looks suspicious. What if the type signature binds some type variables?

Language/Haskell/TH/Desugar/Match.hs
```diff
@@ -142,7 +142,9 @@ tidy1 v (DBangPa pat) =
...
     DWildPa    -> return (id, DBangPa pat)  -- no change
+tidy1 v (DSigPa pat _) = tidy1 v pat
```
goldfirere: Ditto here. I'm dubious of just dropping the signature.

Language/Haskell/TH/Desugar/Match.hs
```diff
@@ -206,6 +208,7 @@ extractBoundNamesDPat (DVarPa n)      = S.singleton n
 extractBoundNamesDPat (DConPa _ pats) = S.unions (map extractBoundNamesDPat pats)
 extractBoundNamesDPat (DTildePa p)    = extractBoundNamesDPat p
 extractBoundNamesDPat (DBangPa p)     = extractBoundNamesDPat p
+extractBoundNamesDPat (DSigPa p _)    = extractBoundNamesDPat p
```
goldfirere: This won't grab bound type variables. Does this matter?

Language/Haskell/TH/Desugar/Match.hs
RyanGlScott: I attempted to model this case off of GHC's [`tidy_bang_pat`](http://git.haskell.org/ghc.git/blob/8f6d241a74efa6f6280689a9b14c36c6a9f4c231:/compiler/deSugar/Match.hs#l482) function, which pays no mind to the type signature. I can understand why you'd be hesitant about this definition, though. I'll add a test which incorporates `ScopedTypeVariables` to stress-test it.

Language/Haskell/TH/Desugar/Match.hs
RyanGlScott: Similarly, I modeled this case off of GHC's [`tidy1`](http://git.haskell.org/ghc.git/blob/8f6d241a74efa6f6280689a9b14c36c6a9f4c231:/compiler/deSugar/Match.hs#l407), which does precisely this.

Language/Haskell/TH/Desugar/Match.hs
RyanGlScott: GHC has a function just like this called [`collectPatBinders`](http://git.haskell.org/ghc.git/blob/8f6d241a74efa6f6280689a9b14c36c6a9f4c231:/compiler/deSugar/DsArrows.hs#l1119) which [also ignores](http://git.haskell.org/ghc.git/blob/8f6d241a74efa6f6280689a9b14c36c6a9f4c231:/compiler/deSugar/DsArrows.hs#l1152) the type in a type pattern, so my hunch is to answer "no, it doesn't matter".

goldfirere:
Glad to see that you based your changes off the code in GHC -- I was not sure of this. With some stress-testing, I'll accept this patch. Thanks!

RyanGlScott:
Currently, this tests [a function](https://github.com/goldfirere/th-desugar/pull/48/files#diff-5fb77260267f093d21a0d7a400cabb69R225) with a scoped type variable that appears in a pattern, as well as a type pattern in [a lambda](https://github.com/goldfirere/th-desugar/pull/48/files#diff-5fb77260267f093d21a0d7a400cabb69R158). Are there other examples you'd like to see?

Test/Splices.hs
```diff
@@ -220,6 +221,11 @@ test40_wildcards = [| let f :: (Show a, _) => a -> a -> _
 test41_typeapps = [| let f :: forall a. (a -> Bool) -> Bool
                          f g = g (undefined @_ @a) in
                      f (const True) |]
+
+test42_scoped_tvs = [| let f :: forall a. Maybe a -> Bool
+                           f (Just (_ :: a)) = True
```
goldfirere: This doesn't really test things, I'm afraid. A pattern signature has funny behavior: if a variable mentioned in a signature is in scope (as this `a` is), then it checks the type. If a variable is not in scope, the signature brings the variable into scope. But since `a` already is in scope, dropping this annotation won't cause trouble. Try something like this:

```
foo :: (Show a, Read a) => a -> String -> String
foo (_ :: b) x = show (read x :: b)
```

This code compiles with the pattern signature but fails without it. Make sure that the test passes through the match-flattener, too.

RyanGlScott:
OK, I've changed the test as per your suggestion. It passes on GHC HEAD, along with every other test (except for the one noted [here](https://github.com/goldfirere/th-desugar/pull/46#issuecomment-266578830), which has nothing to do with this PR).

goldfirere:
In response to failing test 19: No, your suggestion would not be cheating. Just put the kind signatures in manually. th-desugar's reification mechanism can't be expected to insert those kind annotations.

But I'm still incredulous about ScopedTypeVariables. I believe that the test you've written passes, but it critically avoids the match flattener. Put that test in the (embarrassingly short) `simplCaseTests` list in `Splices` and see what happens.

Why am I so worried? GHC's match flattener happens in the desugarer, *after* all the typing information (including scoped type variables) has been processed. So the type annotations are fully redundant there. On the other hand, the output of th-desugar still needs to be type-checked, and so we may have a problem here.

If the match-flattening test fails (as I expect it will), I'm OK with a check in the match-flattener for scoped type variables and simply issuing an error in that unhandled case. (Figuring out how to treat scoped type variables during match flattening is more than we need to consider without an eager client of the feature.)

Thanks!

RyanGlScott:
> In response to failing test 19: No, your suggestion would not be cheating. Just put the kind signatures in manually. th-desugar's reification mechanism can't be expected to insert those kind annotations.

Phew, that's good to know. I've applied that very change.

> But I'm still incredulous about ScopedTypeVariables. I believe that the test you've written passes, but it critically avoids the match flattener. Put that test in the (embarrassingly short) `simplCaseTests` list in `Splices` and see what happens.

Uh-oh, you were right to be skeptical. Here's what get generates (from `-ddump-splices`):

```haskell
     ((let
         f_akFW ::
           forall a_akFV.
           (Read a_akFV, Show a_akFV) => a_akFV -> String -> String
         f_akFW (_ :: b_a9m1) x_akFX = show (read x_akFX :: b_a9m1)
       in f_akFW True "True")
        ==
          (let
             f_akFW ::
               forall a_akFV.
               (Read a_akFV, Show a_akFV) => a_akFV -> String -> String
             f_akFW _arg_1627469450_akGf _arg_1627469452_akGh
               = let x_akFX = _arg_1627469452_akGh in show (read x_akFX :: b_a9m1)
           in f_akFW True "True"))
```

And here's the corresponding error message:

```
Test/Run.hs:236:16: error:
    • Could not deduce (Show a0) arising from a use of ‘show’
      from the context: (Read a, Show a)
        bound by the type signature for:
                   f_akFW :: (Read a, Show a) => a -> String -> String
        at Test/Run.hs:(236,16)-(241,38)
      The type variable ‘a0’ is ambiguous
      These potential instances exist:
        instance [safe] Show Counts -- Defined in ‘Test.HUnit.Base’
        instance [safe] Show Node -- Defined in ‘Test.HUnit.Base’
        instance [safe] Show State -- Defined in ‘Test.HUnit.Base’
        ...plus 98 others
        ...plus 121 instances involving out-of-scope types
        (use -fprint-potential-instances to see them all)
    • In the expression: show (read x_akFX :: b)
      In the expression:
        let x_akFX = _arg_1627469452_akGh in show (read x_akFX :: b)
      In an equation for ‘f_akFW’:
          f_akFW _arg_1627469450_akGf _arg_1627469452_akGh
            = let x_akFX = _arg_1627469452_akGh in show (read x_akFX :: b)

Test/Run.hs:236:16: error:
    • Could not deduce (Read b) arising from a use of ‘read’
      from the context: (Read a, Show a)
        bound by the type signature for:
                   f_akFW :: (Read a, Show a) => a -> String -> String
        at Test/Run.hs:(236,16)-(241,38)
      Possible fix:
        add (Read b) to the context of
          an expression type signature:
            b
    • In the first argument of ‘show’, namely ‘read x_akFX :: b’
      In the expression: show (read x_akFX :: b)
      In the expression:
        let x_akFX = _arg_1627469452_akGh in show (read x_akFX :: b)
```

> If the match-flattening test fails (as I expect it will), I'm OK with a check in the match-flattener for scoped type variables and simply issuing an error in that unhandled case. (Figuring out how to treat scoped type variables during match flattening is more than we need to consider without an eager client of the feature.)

When you say "scoped type variables" here, would you have the match flattener reject _both_ of these programs?

```haskell
foo1 :: forall a. (Show a, Read a) => a -> String -> String
foo1 (_ :: a) x = show (read x :: a)

foo2 :: (Show a, Read a) => a -> String -> String
foo2 (_ :: b) x = show (read x :: b)
```

goldfirere:
Yes, reject both, for simplicity. The first isn't really scoped type variables (in the pattern), but it's very hard to know that. So reject pattern signature that mentions a type variable. Simple!

Thank you so much for all this work!

RyanGlScott:
Very well, I've changed `tidy1` in the match-flattener to unequivocally reject any `DSigPa` types that contain type variables. I've also tweaked the `simpleCaseTests` addition so that it still tests a function with a `DSigPa`, but one that the match-flattener can handle.

goldfirere:
Hooray!

PR #49, opened 2016-12-10
Consolidate duplicate definitions of extractBoundNamesDPat

RyanGlScott:
Currently, `extractBoundNamesDPat` is defined in two different places (but with the same implementation). This PR moves `extractBoundNamesDPat` to `Language.Haskell.TH.Desugar.Core` so that it lives in a central location.


goldfirere:
Did I really do that? Ick. Ick. Thanks for the fix.

PR #50, opened 2016-12-14
Implement unboxed sums

RyanGlScott:
One more TH feature that will be introduced in GHC 8.2.

### DIFF
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -46,6 +46,7 @@ data DExp = DVarE Name
           | DLetE [DLetDec] DExp
           | DSigE DExp DType
           | DStaticE DExp
+          | DUnboxedSumE DExp SumAlt SumArity
           deriving (Show, Typeable, Data, Generic)
 
 
@@ -57,6 +58,7 @@ data DPat = DLitPa Lit
           | DBangPa DPat
           | DSigPa DPat DType
           | DWildPa
+          | DUnboxedSumPa DPat SumAlt SumArity
           deriving (Show, Typeable, Data, Generic)
 
 -- | Corresponds to TH's @Type@ type, used to represent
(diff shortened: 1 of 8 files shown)


Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -70,6 +72,7 @@ data DType = DForallT [DTyVarBndr] DCxt DType
            | DLitT TyLit
            | DWildCardT
            | DStarT
+           | DUnboxedSumT SumArity
```
RyanGlScott: You might be wondering why I didn't try to co-opt `DConT` for this, similarly to how unboxed tuples are currently handled. The reason is that there is currently no good way to represent the `Name`s of prefix unboxed sum type constructors (see [GHC Trac #12514](https://ghc.haskell.org/trac/ghc/ticket/12514)). The same reasoning applies to unboxed sum expressions and patterns. As a result, it was most natural to represent unboxed sums as separate constructors.

goldfirere:
Your comment above is telling. I'm rather against the idea of adding lots of new constructors for this case. Even if you can't write `''(# | #)` or some such, perhaps you can get what you want by using the `Name` datatype from TH concretely. (Yes, I know this is shunned. But it's to work around a GHC infelicity.)

Alternatively, spend your time writing up a ghc-proposal about naming of sum constructors, settle on a direction, implement it, and then return to th-desugar.

RyanGlScott:
That's a very reasonable concern. I think grabbing the `Name` directly in TH might be possible once [Phab:D2854](https://phabricator.haskell.org/D2854) lands.

goldfirere:
Though I don't recommend it, given that you're already putting the fix into template-haskell (where it belongs), you don't have to wait. `Language.Haskell.TH.Syntax` exposes all the bits you need to do this in user land.

RyanGlScott:
@goldfirere: I've updated this PR so that incorporating unboxed sums is a matter of reusing `DCon` from `th-desugar`.

goldfirere:
Looks great. Thanks!

PR #51, opened 2016-12-27
Implement pattern synonyms

RyanGlScott:
This should be the last new feature in GHC HEAD that's currently missing in `th-desugar`.

### DIFF
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -851,7 +851,9 @@ dsDec (RoleAnnotD n roles) = return [DRoleAnnotD n roles]
 #if __GLASGOW_HASKELL__ >= 801
 dsDec (PatSynD n args dir pat) = do
   dir' <- dsPatSynDir n dir
-  (pat', []) <- dsPatX pat
+  (pat', vars) <- dsPatX pat
+  unless (null vars) $
+    fail $ "Pattern synonym definition cannot contain as-patterns (@)."
   return [DPatSynD n args dir' pat']
 dsDec (PatSynSigD n ty) = (:[]) <$> (DPatSynSigD n <$> dsType ty)
 dsDec (StandaloneDerivD mds cxt ty) =
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -127,11 +127,31 @@ data DDec = DLetDec DLetDec
           | DRoleAnnotD Name [Role]
           | DStandaloneDerivD (Maybe DerivStrategy) DCxt DType
           | DDefaultSigD Name DType
+          | DPatSynD Name PatSynArgs DPatSynDir DPat
+          | DPatSynSigD Name DPatSynType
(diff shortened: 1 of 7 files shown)


Language/Haskell/TH/Desugar/Reify.hs
```diff
@@ -240,6 +240,9 @@ reifyInDec n _    dec@(ClosedTypeFamilyD n' _ _ _) | n `nameMatches` n'
   = Just $ FamilyI dec []
 #endif
 #endif
+#if __GLASGOW_HASKELL__ >= 801
+reifyInDec n _ (PatSynSigD n' ty) | n `nameMatches` n' = Just $ PatSynI n ty
```
RyanGlScott: I had no clue what I was doing when implementing this particular case, so please scrutinize this carefully.

goldfirere: Echo the design for `mkVarI`: find whether or not a pattern synonym of the right name is declared, and if so, search for the signature. If the synonym is declared but the signature is missing, report an error, as reification morally should work, but th-desugar does not do type inference.

Language/Haskell/TH/Desugar/Sweeten.hs
```diff
@@ -172,6 +175,10 @@ decToTH (DStandaloneDerivD _mds cxt ty) =
     (cxtToTH cxt) (typeToTH ty)]
 decToTH (DDefaultSigD n ty)        = [DefaultSigD n (typeToTH ty)]
 #endif
+#if __GLASGOW_HASKELL__ >= 801
+decToTH (DPatSynD n args dir pat) = [PatSynD n args (patSynDirToTH dir) (patToTH pat)]
+decToTH (DPatSynSigD n ty)        = [PatSynSigD n (typeToTH ty)]
+#endif
```
goldfirere: This needs an `#else`. Otherwise you'll get the "newtype" error below.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -825,6 +849,11 @@ dsDec (RoleAnnotD n roles) = return [DRoleAnnotD n roles]
 #endif
 #if __GLASGOW_HASKELL__ >= 709
 #if __GLASGOW_HASKELL__ >= 801
+dsDec (PatSynD n args dir pat) = do
+  dir' <- dsPatSynDir n dir
+  (pat', []) <- dsPatX pat
```
RyanGlScott: Ah, I stubbed that out and forgot about the irrefutable pattern match.

In any case, [as-patterns are disallowed in pattern synonyms](http://git.haskell.org/ghc.git/blob/a3704409acc3bd237d3e872f640686918fb51f5f:/compiler/typecheck/TcPatSyn.hs#l749) (and this is [by design](http://git.haskell.org/ghc.git/blob/a3704409acc3bd237d3e872f640686918fb51f5f:/compiler/typecheck/TcPatSyn.hs#l695)). But I've added a separate check here as well just to be safe.

goldfirere:
Looks great. Thanks for the updates. Will merge once Travis says OK.

goldfirere:
Thanks!

PR #52, opened 2017-01-04
Cache dependencies in Travis builds

RyanGlScott:
About half of the time in the Travis build is spend just building dependencies, so let's cache them. This adapts the `.travis.yml` script to a more modern version that:

1. Caches dependencies
2. Uses containerized Travis builds (which are apparently [faster](https://docs.travis-ci.com/user/migrating-from-legacy/#Why-migrate-to-container-based-infrastructure%3F) than non-containerized ones)
3. Adds a GHC 8.0.2 build

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,43 +1,98 @@
-# NB: don't set `language: haskell` here
+# This file has been generated -- see https://github.com/hvr/multi-ghc-travis
+language: c
+sudo: false
 
-# explicitly request legacy non-sudo based build environment
-sudo: required
+cache:
+  directories:
+    - $HOME/.cabsnap
+    - $HOME/.cabal/packages
 
-# The following enables several GHC versions to be tested; often it's enough to test only against the last release in a major GHC version. Feel free to omit lines listings versions you don't need/want testing for.
-env:
- - CABALVER=1.24 GHCVER=7.6.3
- - CABALVER=1.18 GHCVER=7.8.2
- - CABALVER=1.18 GHCVER=7.8.3
- - CABALVER=1.18 GHCVER=7.8.4
- - CABALVER=1.22 GHCVER=7.10.1
- - CABALVER=1.22 GHCVER=7.10.2
- - CABALVER=1.22 GHCVER=7.10.3
- - CABALVER=1.24 GHCVER=8.0.1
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,43 +1,98 @@
...
+      addons: {apt: {packages: [cabal-install-1.18,ghc-7.8.3], sources: [hvr-ghc]}}
+    - env: CABALVER=1.18 GHCVER=7.8.4
+      compiler: ": #GHC 7.8.4"
+      addons: {apt: {packages: [cabal-install-1.18,ghc-7.8.4], sources: [hvr-ghc]}}
+    - env: CABALVER=1.22 GHCVER=7.10.1
+      compiler: ": #GHC 7.10.1"
+      addons: {apt: {packages: [cabal-install-1.22,ghc-7.10.1], sources: [hvr-ghc]}}
+    - env: CABALVER=1.22 GHCVER=7.10.2
+      compiler: ": #GHC 7.10.2"
+      addons: {apt: {packages: [cabal-install-1.22,ghc-7.10.2], sources: [hvr-ghc]}}
```
RyanGlScott: What's your policy on testing old point releases like GHC 7.10.1 and 7.10.2? I kept them in since they were present in the previous iteration of `.travis.yml`, but is it sufficient to just test GHC 7.10.3 from here on out?

.travis.yml
```diff
@@ -47,5 +102,4 @@ script:
  - SRC_TGZ=$(cabal info . | awk '{print $2;exit}').tar.gz &&
    (cd dist && cabal install --force-reinstalls "$SRC_TGZ")
 
-after_script:
-  - cat dist/test/*.log
```
RyanGlScott: I removed this part since it's superseded by the `--show-details=always` flag of `cabal test`.

.travis.yml
goldfirere: For th-desugar, I'd like to test the point releases, as cross-version compatibility for TH is one of the goals of the library. Not to say that we'd write tons of bespoke code for a point release, but we should at least learn of the breakage.

goldfirere:
Thanks for doing this. Updating Travis has been something on my todo list, but very low on that list.

PR #54, opened 2017-02-12
Fix error on pragmas in 'let'-declarations

DanielSchuessler:
Hello,

this currently throws an error:
```haskell
test44_let_pragma = [| let x :: Int
                           x = 1
                           {-# INLINE x #-}
                       in x |]
```
I fixed it by moving `DPragmaD` into `DLetDec`. Not sure how much of an issue backwards compatibility is for this package; maybe `DPragmaD` could be made a pattern synonym and the real constructor in `DLetDec` could be named differently.

Another unfortunate breaking change is that `letDecToTH` has to return a `Maybe` now due to the possibility of unsupported pragmas.

Not sure about recursing into pragmas in `scLetDec` (`DExp`s in rules and annotations), but I did so since `scLetExp` mentions a "guarantee".

### DIFF
--- a/Language/Haskell/TH/Desugar/Sweeten.hs
+++ b/Language/Haskell/TH/Desugar/Sweeten.hs
@@ -51,10 +51,7 @@ expToTH (DLitE l)            = LitE l
 expToTH (DAppE e1 e2)        = AppE (expToTH e1) (expToTH e2)
 expToTH (DLamE names exp)    = LamE (map VarP names) (expToTH exp)
 expToTH (DCaseE exp matches) = CaseE (expToTH exp) (map matchToTH matches)
-expToTH (DLetE decs exp)     = case mapMaybe letDecToTH decs of
-                                  -- This can only happen if we somehow have a DLetE containing only pragmas
-                                  [] -> expToTH exp
-                                  decs' -> LetE decs' (expToTH exp)
+expToTH (DLetE decs exp)     = LetE (mapMaybe letDecToTH decs) (expToTH exp)
 expToTH (DSigE exp ty)       = SigE (expToTH exp) (typeToTH ty)
 #if __GLASGOW_HASKELL__ < 709
 expToTH (DStaticE _)         = error "Static expressions supported only in GHC 7.10+"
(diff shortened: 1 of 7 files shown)


RyanGlScott:
For the sake of being explicit, here is what "an error" entails in this case:

```
λ> $([| let x :: Int; x = 1; {-# INLINE x #-} in x |] >>= desugar >>= stringE . show . expToTH)

<interactive>:9:3: error:
    • Illegal declaration in let expression.
    This should not happen in Haskell.
    Please email eir@cis.upenn.edu with your code if you see this.
    • In the untyped splice:
        $([| let
               x :: Int
               x = 1
               ....
             in x |]
          >>= desugar
          >>= stringE
          . show
          . expToTH)
```

Language/Haskell/TH/Desugar/Sweeten.hs
```diff
@@ -51,7 +51,10 @@ expToTH (DLitE l)            = LitE l
 expToTH (DAppE e1 e2)        = AppE (expToTH e1) (expToTH e2)
 expToTH (DLamE names exp)    = LamE (map VarP names) (expToTH exp)
 expToTH (DCaseE exp matches) = CaseE (expToTH exp) (map matchToTH matches)
-expToTH (DLetE decs exp)     = LetE (map letDecToTH decs) (expToTH exp)
+expToTH (DLetE decs exp)     = case mapMaybe letDecToTH decs of
+                                  -- This can only happen if we somehow have a DLetE containing only pragmas
+                                  [] -> expToTH exp
```
goldfirere: I don't think this needs to be special-cased. You might end up with a `LetE` with no declarations, but such a thing is OK.

DanielSchuessler:
Right, I forgot empty `LetE`s are allowed. I guess `expToTH` is intended to do only a literal translation, no prettification? (Such as producing `TupE`s instead of `AppE`s of the tuple constructor, to make the result easier to read when debugging. Not that it's very important :))

Looks like `safe`'s build failure on GHC 7.10.1 is a known issue: https://github.com/ndmitchell/safe/issues/20#issuecomment-278955186 . I added it to `allow_failures` in the .travis.yml.

RyanGlScott:
FWIW, https://github.com/ndmitchell/safe/issues/20 was fixed with the release of [`safe-0.3.14`](http://hackage.haskell.org/package/safe-0.3.14), so you should be able to remove GHC 7.10.1 from the `allow_failures` list (I think).

goldfirere:
Oops. I guess I let this one fall through the cracks. Merging, but we should fix the .travis.yml file as Ryan suggests.

PR #55, opened 2017-04-08
Fix getRecordSelectors at the source

RyanGlScott:
This migrates the changes I originally put forth in https://github.com/goldfirere/singletons/pull/181 to the site of `getRecordSelectors`' definition. `getRecordSelectors` now requires a _list_ of `DCons` instead of just one, which should make it harder to trigger the bug in https://github.com/goldfirere/singletons/issues/180.

### DIFF
--- a/Language/Haskell/TH/Desugar.hs
+++ b/Language/Haskell/TH/Desugar.hs
@@ -5,7 +5,7 @@ eir@cis.upenn.edu
 -}
 
 {-# LANGUAGE CPP, MultiParamTypeClasses, FunctionalDependencies,
-             TypeSynonymInstances, FlexibleInstances #-}
+             TypeSynonymInstances, FlexibleInstances, LambdaCase #-}
 
 -----------------------------------------------------------------------------
 -- |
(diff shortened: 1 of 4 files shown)


Language/Haskell/TH/Desugar.hs
```diff
@@ -188,30 +188,74 @@ dtvbName (DPlainTV n)    = S.singleton n
...
-        ]
+    get_record_sels (DCon _ _ con_name con _) = case con of
+      DRecC fields -> go fields
+      _ -> return []
+      where
+        go fields = do
+          varName <- qNewName "field"
+          let tvbs = fvDType arg_ty
+              maybe_forall
+                | S.null tvbs = id
```
goldfirere: Why the special case? An empty `forall` is OK.

goldfirere: OK. Now I realize that you're just copying my code. But I still think an empty `forall` is OK.

Language/Haskell/TH/Desugar.hs
```diff
@@ -188,30 +188,74 @@ dtvbName (DPlainTV n)    = S.singleton n
...
     mk_field_pats n total name = DWildPa : mk_field_pats (n-1) (total-1) name
+
+    merge_let_decs [] = []
+    merge_let_decs (x:xs)
+      -- If we encounter a function declaration, looks for all other function
+      -- declarations in the rest of the list with the same name, and concat
+      -- their clauses.
+      | DFunD n clauses <- x
+      = let (other_clauses, neq_n)
+              = partitionWith (\case DFunD n2 cls | n == n2 -> Left cls
```
goldfirere: This is quadratic. :( Would it work to use a finite map (that is, `Data.Map`) mapping names to lists of clauses? That seems much more efficient.

Language/Haskell/TH/Desugar/Util.hs
```diff
@@ -273,6 +273,15 @@ mapMaybeM f (x:xs) = do
     Nothing -> ys
     Just z  -> z : ys
 
+-- like GHC's
+partitionWith :: (a -> Either b c) -> [a] -> ([b], [c])
```
goldfirere: It's a shame this isn't in some library. Very useful function.

RyanGlScott:
@goldfirere, thanks for the review. I've updated the internals of `getRecordSelectors` so that it isn't quadratic anymore, and rejigged the documentation appropriately. Does its new incarnation look OK?

Language/Haskell/TH/Desugar.hs
```diff
@@ -188,30 +189,112 @@ dtvbName (DPlainTV n)    = S.singleton n
...
+                    -> (M.Map Name [DClause], [DLetDec])
+        gather_decs name_clause_map _ [] = (name_clause_map, [])
+        gather_decs name_clause_map type_sig_names (x:xs)
+          -- 1.
+          | DFunD n clauses <- x
+          = if n `M.member` name_clause_map
+            then gather_decs (M.insertWith (\new old -> old ++ new)
+                                           n clauses name_clause_map)
+                             type_sig_names xs
+            else let (map', decs') = gather_decs (M.insert n clauses name_clause_map)
```
goldfirere: The previous call to `gather_decs` will work here, right? If so, perhaps bind that result somewhere, so it doesn't look like the map-insertion is changing between the `then` and `else`.

PR #58, opened 2017-06-16
Don't crash when expanding quasiquoted type families

RyanGlScott:
I discovered this when working on https://github.com/goldfirere/singletons/issues/198. If you compile this with `th-desugar-1.7`:

```haskell
module Bug where

import Language.Haskell.TH
import Language.Haskell.TH.Desugar.Lift

test_local_tyfam_expansion :: Bool
test_local_tyfam_expansion =
  $(do fam_name <- newName "Fam"
       let orig_ty = DConT fam_name
       exp_ty <- withLocalDeclarations
                   (decsToTH [ DOpenTypeFamilyD (DTypeFamilyHead fam_name [] DNoSig Nothing)
                             , DTySynInstD fam_name (DTySynEqn [] (DConT ''Int)) ])
                   (expandType orig_ty)
       [| orig_ty == exp_ty |])
```

It'll break:

```
GHCi, version 8.2.0.20170523: http://www.haskell.org/ghc/  :? for help

Bug.hs:10:5: error:
    • The exact Name ‘Fam_a7Co’ is not in scope
        Probable cause: you used a unique Template Haskell name (NameU), 
        perhaps via newName, but did not bind it
        If that's it, then -ddump-splices might be useful
    • In the argument of reifyInstances: Fam_0
      In the untyped splice:
        $(do fam_name <- newName "Fam"
             let orig_ty = DConT fam_name
             exp_ty <- withLocalDeclarations
                         (decsToTH
                            [DOpenTypeFamilyD (DTypeFamilyHead fam_name [] DNoSig Nothing),
                             DTySynInstD fam_name (DTySynEqn [] (DConT ''Int))])
                         (expandType orig_ty)
             [| orig_ty == exp_ty |])
   |
10 |   $(do fam_name <- newName "Fam"
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^...

Bug.hs:10:5: error:
    • GHC internal error: ‘Fam_a7Co’ is not in scope during type checking, but it passed the renamer
      tcl_env of environment: []
    • In the type ‘Fam_a7Co’
      In the argument of reifyInstances: Fam_0
      In the untyped splice:
        $(do fam_name <- newName "Fam"
             let orig_ty = DConT fam_name
             exp_ty <- withLocalDeclarations
                         (decsToTH
                            [DOpenTypeFamilyD (DTypeFamilyHead fam_name [] DNoSig Nothing),
                             DTySynInstD fam_name (DTySynEqn [] (DConT ''Int))])
                         (expandType orig_ty)
             [| orig_ty == exp_ty |])
   |
10 |   $(do fam_name <- newName "Fam"
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

That GHC internal error is due to [Trac #13837](https://ghc.haskell.org/trac/ghc/ticket/13837), but that's a separate problem. The more pressing issue is that `reifyInstances` is failing at all. This happens because a use of `qReifyInstances` in `expand_con` fails when a type family is declared locally (e.g., in quasiquotes).

This PR simply guards that use of `qReifyInstances` with `qRecover`. In an ideal world, we'd have `qReifyInstancesWithLocals`, but that would require implementing instance lookup in `th-desugar`, which sounds far above my paygrade. This is a much simpler solution.

### DIFF
--- a/Language/Haskell/TH/Desugar/Expand.hs
+++ b/Language/Haskell/TH/Desugar/Expand.hs
@@ -118,8 +118,11 @@ expand_con ign n args = do
 #endif
       -> do
         let (syn_args, rest_args) = splitAtList tvbs args
-        -- need to get the correct instance
-        insts <- qReifyInstances n (map typeToTH syn_args)
+        -- We need to get the correct instance. If we fail to reify anything
+        -- (e.g., if a type family is quasiquoted), then fall back by
+        -- pretending that there are no instances in scope.
+        insts <- qRecover (return []) $
+                 qReifyInstances n (map typeToTH syn_args)
         dinsts <- dsDecs insts
         case dinsts of
           [DTySynInstD _n (DTySynEqn lhs rhs)] -> do
--- a/Test/Run.hs
+++ b/Test/Run.hs
@@ -228,6 +228,16 @@ test_deriving_strategies = compare (MkBlarggie 5 'x') (MkBlarggie 5 'x') == EQ
 test_deriving_strategies = True
 #endif
 
+test_local_tyfam_expansion :: Bool
+test_local_tyfam_expansion =
+  $(do fam_name <- newName "Fam"
+       let orig_ty = DConT fam_name
+       exp_ty <- withLocalDeclarations
+                   (decsToTH [ DOpenTypeFamilyD (DTypeFamilyHead fam_name [] DNoSig Nothing)
+                             , DTySynInstD fam_name (DTySynEqn [] (DConT ''Int)) ])
+                   (expandType orig_ty)
+       orig_ty `eqTHSplice` exp_ty)
+
 local_reifications :: [String]
 local_reifications = $(do decs <- reifyDecs
                           m_infos <- withLocalDeclarations decs $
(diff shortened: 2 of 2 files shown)


goldfirere:
Agreed that we don't want to do instance lookup ourselves.... but should we issue a warning? If someone calls `reifyInstances` on a type for which we can't retrieve instances, I feel like we owe them some explanation of what's going on.

RyanGlScott:
I'm not sure if would be wise to issue a warning here. The problem is that you might be in a scenario where a type family has no instances:

```haskell
module Bug where

import Language.Haskell.TH
import Language.Haskell.TH.Desugar.Lift

test_local_tyfam_expansion :: Bool
test_local_tyfam_expansion =
  $(do fam_name <- newName "Fam"
       let orig_ty = DConT fam_name
       exp_ty <- withLocalDeclarations
                   (decsToTH [ DOpenTypeFamilyD (DTypeFamilyHead fam_name [] DNoSig Nothing) ])
                   (expandType orig_ty)
       [| orig_ty == exp_ty |])
```

Here, you still need to call `qReifyInstances` anyway during `expandType`, so it's still subject to the same bug that this PR fixed. I would be rather annoyed if `th-desugar` kept pestering me that `qReifyInstances` couldn't find any instances for `Fam` even though I know it wasn't going to find anything in the first place.

Moreover, in goldfirere/singletons#198, the primary use case for this feature, `th-desugar` would give you a warning every time you used a type family within `$(singletons [d| ... |])`, which feels horribly gross to me.

goldfirere:
Grumble. OK.

PR #62, opened 2017-07-12
Implement lookup{Value,Type}NameWithLocals

RyanGlScott:
I ended up needing versions of `lookupValueName` and `lookupTypeName` that work with local declarations while working on https://github.com/goldfirere/singletons/issues/163, so this PR adds them to `th-desugar`.

This diff looks busy, but most of the changes involve plumbing up the looked up `Name` so that it can be quickly retrieved later (instead of having to re-pattern match the `Info` we get from `reifyInDecs` in order to get out the `Name`).

### DIFF
--- a/Language/Haskell/TH/Desugar/Util.hs
+++ b/Language/Haskell/TH/Desugar/Util.hs
@@ -48,6 +48,11 @@ newUniqueName str = do
   n <- qNewName str
   qNewName $ show n
 
+-- | @mkNameWith lookup_fun mkName_fun str@ looks up the exact 'Name' of @str@
+-- using the function @lookup_fun@. If it finds 'Just' the 'Name', meaning
+-- that it is bound in the current scope, then it is returned. If it finds
+-- 'Nothing', it assumes that @str@ is declared in the current module, and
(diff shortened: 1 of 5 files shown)


RyanGlScott:
I just noticed that `th-desugar` also defines `mkDataName` and `mkTypeName` functions, which use `qLookupName` under the hood. We should probably also have `mkDataNameWithLocals` and `mkTypeNameWithLocals` for completeness.

Language/Haskell/TH/Desugar/Util.hs
```diff
@@ -48,27 +48,26 @@ newUniqueName str = do
   n <- qNewName str
   qNewName $ show n
 
--- | Like TH's @lookupTypeName@, but if this name is not bound, then we assume
--- it is declared in the current module.
-mkTypeName :: Quasi q => String -> q Name
-mkTypeName str = do
-  m_name <- qLookupName True str
+mkNameWith :: Quasi q => (String -> q (Maybe Name))
```
goldfirere: Documentation? That type is intricate enough that this wants a few lines of comments.

PR #64, opened 2017-07-26
Implement reifyNameSpace

RyanGlScott:
This ends up being pretty simple, thanks to all of the existing machinery in `Language.Haskell.TH.Desugar.Reify`.

### DIFF
--- a/Language/Haskell/TH/Desugar/Reify.hs
+++ b/Language/Haskell/TH/Desugar/Reify.hs
@@ -22,6 +22,7 @@ module Language.Haskell.TH.Desugar.Reify (
   -- * Value and type lookup
   lookupValueNameWithLocals, lookupTypeNameWithLocals,
   mkDataNameWithLocals, mkTypeNameWithLocals,
+  reifyNameSpace,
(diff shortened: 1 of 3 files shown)


Language/Haskell/TH/Desugar/Reify.hs
```diff
@@ -615,3 +609,32 @@ mkDataNameWithLocals = mkNameWith lookupValueNameWithLocals mkNameG_d
 -- determining if the name is currently bound.
 mkTypeNameWithLocals :: DsMonad q => String -> q Name
 mkTypeNameWithLocals = mkNameWith lookupTypeNameWithLocals mkNameG_tc
+
+reifyNameSpace :: DsMonad q => Name -> q (Maybe NameSpace)
```
goldfirere: This needs a Haddock comment, explaining (among other things) what happens if the identifier is in both namespaces.

RyanGlScott: Oops, I can't believe I forgot that. Added.

PR #65, opened 2017-08-03
Add (Maybe DKind) field to DDataD/DDataFamilyD/DDataInstD

RyanGlScott:
The inability to express data types and data families with explicit return kinds prevents one from using `th-desugar` to construct certain data type signatures (e.g., `data Foo :: forall (a :: k). Proxy a -> Type`, or what is requested in https://github.com/goldfirere/singletons/issues/216#issuecomment-320037390), so this PR adds that ability.

Previously, we were awkwardly converting the return kinds of data types and data families into `DTyVarBndr`s. Now that we have explicit fields for return kinds, this is no longer necessary, allowing us to remove the dreaded `mkExtraTvbs` function.

There were already a number of tests that reify datatypes with return kinds like `:: * -> *`, so I opted not to clutter the test suite with more tests of that ilk. (I would test quoting something exotic like `data Foo :: forall k. k -> Type`, but that capability is only present in GHC HEAD, so I decided it wouldn't be worth it.)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,10 @@ Version 1.8
 * Incorporate a `DDeclaredInfix` field into `DNormalC` to indicate if it is
   a constructor that was declared infix.
 
+* Incorporate a `Maybe DKind` field into `DDataD`, `DDataFamilyD`, and
+  `DDataInstD` to reflect the return kind of a datatype or data family (if it
+  was declared with an explicit one).
+
 * Implement `lookupValueNameWithLocals`, `lookupTypeNameWithLocals`,
   `mkDataNameWithLocals`, and `mkTypeNameWithLocals`, counterparts to
   `lookupValueName`, `lookupTypeName`, `mkDataName`, and `mkTypeName` which
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -16,7 +16,6 @@ import Prelude hiding (mapM, foldl, foldr, all, elem, exp, concatMap, and)
 
 import Language.Haskell.TH hiding (match, clause, cxt)
 import Language.Haskell.TH.Syntax hiding (lift)
-import Language.Haskell.TH.ExpandSyns ( expandSyns )
 
 #if __GLASGOW_HASKELL__ < 709
 import Control.Applicative
@@ -117,15 +116,15 @@ data NewOrData = Newtype
 
 -- | Corresponds to TH's @Dec@ type.
 data DDec = DLetDec DLetDec
-          | DDataD NewOrData DCxt Name [DTyVarBndr] [DCon] [DDerivClause]
+          | DDataD NewOrData DCxt Name [DTyVarBndr] (Maybe DKind) [DCon] [DDerivClause]
           | DTySynD Name [DTyVarBndr] DType
           | DClassD DCxt Name [DTyVarBndr] [FunDep] [DDec]
           | DInstanceD (Maybe Overlap) DCxt DType [DDec]
           | DForeignD DForeign
           | DOpenTypeFamilyD DTypeFamilyHead
           | DClosedTypeFamilyD DTypeFamilyHead [DTySynEqn]
-          | DDataFamilyD Name [DTyVarBndr]
-          | DDataInstD NewOrData DCxt Name [DType] [DCon] [DDerivClause]
+          | DDataFamilyD Name [DTyVarBndr] (Maybe DKind)
+          | DDataInstD NewOrData DCxt Name [DType] (Maybe DKind) [DCon] [DDerivClause]
           | DTySynInstD Name DTySynEqn
           | DRoleAnnotD Name [Role]
           | DStandaloneDerivD (Maybe DerivStrategy) DCxt DType
@@ -820,7 +819,7 @@ fixBug8884ForFamilies (DClosedTypeFamilyD (DTypeFamilyHead name tvbs frs ann) eq
       eqns' = map (fixBug8884ForEqn num_args) eqns
   frs' <- remove_arrows num_args frs
   return (DClosedTypeFamilyD (DTypeFamilyHead name tvbs frs' ann) eqns', num_args)
-fixBug8884ForFamilies dec@(DDataFamilyD _ _)
+fixBug8884ForFamilies dec@(DDataFamilyD{})
   = return (dec, 0)   -- the num_args is ignored for data families
 fixBug8884ForFamilies dec =
   impossible $ "Reifying yielded a FamilyI with a non-family Dec: " ++ show dec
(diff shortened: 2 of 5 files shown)


goldfirere:
There is a design choice to be considered here. When initially envisioned, the purpose of th-desugar was to support the smallest possible AST that could capture all the semantics of GHC. There are a few areas where th-desugar falls a bit short here, but that's been the goal. This patch takes a step backwards, given that `data X :: Type -> Type` is fully semantically equivalent to `data X (a :: Type)`.

Of course, this move is in service of a worthy goal: make it easier to generate syntax. But I wonder if it makes more sense just to generate TH AST directly -- th-desugar is not meant to be generated; it's meant to be consumed. What would it look like if, say, `singletons` just generated TH directly? I don't actually imagine that would be harder at all! After all, th-desugar is pretty much a subset of TH. And doing so would narrow th-desugar's charter, in what would seem a helpful direction.

RyanGlScott:
The problem is that, AFAICT, there are some return kinds that don't have straightforward counterparts in the tyvar-binder-only style that `th-desugar` supports. For instance, how would `th-desugar` represent this?

```haskell
data Foo :: forall j. j -> forall k. k -> Type
```

goldfirere:
Good point. I'm still a bit bothered by the way the th-desugar AST is growing, though. While we don't have to go here today, perhaps this would be a nice long-term goal: separate out GADT definitions from regular ones. The AST in this PR allows a non-GADT definition to have a kind signature, disallowed in regular Haskell. If we separated out GADT definitions from H98 definitions, then we could *require* a kind signature on GADT definitions instead of having tyvars. And we'd be back to something resembling a minimal AST.

RyanGlScott:
> The AST in this PR allows a non-GADT definition to have a kind signature, disallowed in regular Haskell.

That's true. But for what it's worth, it's possible today to use `th-desugar` to embed GADT properties for non-GADTs in other places, too. Namely, in `DCon`, which has a `Maybe DType` that is only `Just` if there's a GADT return type. It would be quite easy to screw with `th-desugar` by having some `DCon`s have `Just` GADT return types and others have `Nothing`.

One solution to this problem would be to conjure up a different datatype:

```haskell
data DataSyntax = H98 | Gadt
```

Now we can redesign `DDataD`, `DDataInstD`, and `DCon` like so:

```haskell
data DDec
 = ...
 | forall (ds :: DataSyntax). DDataD NewOrData DCxt Name [DTyVarBndr] (GadtKind ds) [DCon ds] [DPred]
 | forall (ds :: DataSyntax). DDataInstD NewOrData DCxt Name [DType] (GadtKind ds) [DCon ds] [DPred]

data DCon (ds :: DataSyntax) = DCon [DTyVarBndr] DCxt Name DConFields (GadtType ds)

data GadtType (ds :: DataSyntax) where
  NotGadt :: GadtType 'H98
  GadtType :: Maybe DType -> GadtType 'Gadt

type GadtKind = GadtType
```

This would require some more GHC extensions, but I think it would make it clearer where GADT syntax is being used.

goldfirere:
Ooh. I like.

RyanGlScott:
Oh crud, there's a big problem with that design: `DDec` would become a GADT, and thus we'd have to give up `Data` and `Generic` instances for a large swath of datatypes in `Language.Haskell.TH.Desugar`. I'm not sure if that would be worth it...

RyanGlScott:
Well, "large swaths" is a bit of an exaggeration. It would just be limited to `DInfo` and `DDec`, I _think_.

RyanGlScott:
If going with GADTs in the `th-desugar` AST is untenable, it might just be worth documenting the invariants that are expected to hold.

goldfirere:
We'd have to give up on `Generic` for those types, yes. `Data`, not so much. `Data` gives us two ways of using a type: consuming it and producing it. Producing a GADT with `Data` is nigh impossible, but *consuming* it is just fine. (Really, there should be two classes to make this distinction clearer.) So, we could potentially hand-write `Data` instances that allow `gfoldl` to work while erroring on `gunfold`. Is this worth it? I'm not sure.

Here's another crazy thought: the reason I care about this is that I want to be able to pattern-match on these types and know I've handled all cases. Instead of using GADTs to get what we want, can we use `COMPLETE` pragmas? That is, teach the pattern-match-checker that non-GADT types can't have GADT constructors. Pattern synonyms might be involved. I haven't worked out any details. But this might be a different way to skin this cat.

RyanGlScott:
If pattern synonyms are fair game, one could try this:

```haskell
data DDec
  = ...
  | DDataD       NewOrData DCxt Name [DTyVarBndr] GadtKind [DCon] [DDerivClause]
  | DDataFamilyD NewOrData DCxt Name [DType]      GadtKind [DCon] [DDerivClause]

data DCon = DCon [DTyVarBndr] DCxt Name DConFields GadtType

type GadtType = Maybe DType
type GadtKind = GadtType

pattern NotGadt :: GadtType
pattern NotGadt = Nothing

pattern GadtType :: DType -> GadtType
pattern GadtType t = Just t
{-# COMPLETE NotGadt  #-}
{-# COMPLETE GadtType #-}
```

This allows the following definitions to compile with no warnings under `-Wall`:

```haskell
h98NamesOfCons :: DDec -> [Name]
h98NamesOfCons (DDataD       _ _ _ _ NotGadt cons _) = map h98NameOfCon cons
h98NamesOfCons (DDataFamilyD _ _ _ _ NotGadt cons _) = map h98NameOfCon cons

h98NameOfCon :: DCon -> Name
h98NameOfCon (DCon _ _ n _ NotGadt) = n

gadtReturnTypes :: DDec -> [DType]
gadtReturnTypes (DDataD       _ _ _ _ GadtType{} cons _) = map gadtReturnType cons
gadtReturnTypes (DDataFamilyD _ _ _ _ GadtType{} cons _) = map gadtReturnType cons

gadtReturnType :: DCon -> DType
gadtReturnType (DCon _ _ _ _ (GadtType t)) = t
```

Would that work for you?

goldfirere:
If those definitions really compile, then `COMPLETE` is terribly broken. What about all the other `DDec`s?

RyanGlScott:
I now realize that my example is a bit misleading. To test out this design, I reduced `DDec` down to just the two datatype-related constructors, which is why I omitted them with an ellipsis (I now seee that one could interpret that as meaning "the other constructors are still there, but I didn't type them out." My bad.)

Indeed, in the presence of the other `DDec` constructors, you would have to include them to be exhaustive.

goldfirere:
This still isn't quite right. If I'm matching against a `DDec`, then it might include a GADT and it might include an H98 declaration. This is not accounted for in your design.

What about yet a different approach: have th-desugar support only GADT constructors / declarations; translate H98 declarations to the GADT style. This translation should be easy enough, and I believe all H98 expressiveness is available in GADT forms. What do you think?

RyanGlScott:
> This still isn't quite right. If I'm matching against a `DDec`, then it might include a GADT and it might include an H98 declaration. This is not accounted for in your design.

Again, I think I chose my example poorly, since it didn't convey my intent. The idea is that (as an example) `h98NamesOfCons` is a function one can use when you know that the invariant that you're only working over H98 datatypes holds (which would otherwise be enforced by `DDec` being an actual GADT, but we had to get creative here). You could just as well make `h98NamesOfCons` work over either H98 datatypes or GADTs, and the `GadtKind` field would make it easy to tell which is which, depending on whether you match on `Nothing` or `Just` (the pattern synonyms are only intended to be used when you know the invariant holds, not all the time).

> What about yet a different approach: have th-desugar support only GADT constructors / declarations; translate H98 declarations to the GADT style. This translation should be easy enough, and I believe all H98 expressiveness is available in GADT forms. What do you think?

There'd be one thorny issue with this: how do we sweeten a `th-desugar` GADT? In GHC 8.0 or later, we could always sweeten it back to a `(Rec)GadtC`, but that wasn't available in versions prior. It doesn't seem right to just fail to fail to sweeten it on old GHCs, but if we sweetened it to, say, an H98 datatype using `ForallC`, then we'd have inconsistencies across GHC versions. Perhaps this suggests we should record whether a constructor was declared using H98 or GADT syntax?

goldfirere:
I'm OK with sweetener inconsistencies across GHC versions. I'm not OK with desugaring inconsistencies, though.... but we wouldn't have that here.

And I understand your first solution better now.... but it still seems quite challenging to use: you say `Nothing` to *check* whether a data declaration is H98 but `NotGadt` when you already know it. Very subtle.

RyanGlScott:
> And I understand your first solution better now.... but it still seems quite challenging to use: you say `Nothing` to check whether a data declaration is H98 but `NotGadt` when you already know it. Very subtle.

Indeed. To be clear: I'm not particularly fond of solving this problem with pattern synonyms, but I sketched it out anyway, since you requested it :)

> I'm OK with sweetener inconsistencies across GHC versions. I'm not OK with desugaring inconsistencies, though.... but we wouldn't have that here.

Thanks for mentioning that, because I was falsely under the impression `th-desugar` was consistently sweetening things across multiple GHC versions, but that's [not true](https://github.com/goldfirere/th-desugar/blob/d3f5c493d037c74c1326c867edac4ee4dd45efe3/Language/Haskell/TH/Desugar/Sweeten.hs#L218-L247).

And now that I think about it, adding an extra field stating whether a constructor was declared using GADT syntax wouldn't even address the issue of sweetening inconsistencies in the first place, since that wouldn't solve the pre- and post-GHC 8.0 Template Haskell design differences. It //would// make it so that more H98 datatypes could roundtrip successfully through `th-desugar`, but if that's not a goal we care about, then it's probably wasted effort.

RyanGlScott:
Now you've convinced me that always desugaring down to GADT syntax is the right way to go. Given that, this PR is obsolete, so I'll close this in favor of a different one.

PR #70, opened 2018-01-26
Remove DStarT

### LINKED ISSUE #69: Remove DStarT?
Currently, `DType` has a distinguished constructor for `*` called `DStarT`. But this is a somewhat questionable choice, given that:

* We could just as easily encode this with `DConT (mkName "*")` (or `DConT ''Type`, on more recent GHCs).
* We are moving in the direction of [embracing `Type` over `*`](https://github.com/ghc-proposals/ghc-proposals/pull/83) in the (perhaps not-so-distant) future.

In light of this, perhaps we ought to simply just remove `DStarT`. This should be a relatively straightforward change save for one complication: we cannot sweeten `DConT (mkName "*")` back to `ConT (mkName "*")` on GHC 7.6 due to an [old bug](https://ghc.haskell.org/trac/ghc/ticket/7667), as this program witnesses:

```haskell
{-# LANGUAGE KindSignatures #-}
module Bug where

import Language.Haskell.TH.Syntax

$(let starKindName :: Name
      starKindName = mkNameG_tc "ghc-prim" "GHC.Prim" "*"
   in do foo <- newName "foo"
         return [ SigD foo (SigT (ConT ''Bool) (ConT starKindName))
                , ValD (VarP foo) (NormalB (ConE 'True)) []
                ])
```
```
$ /opt/ghc/7.6.3/bin/ghc Bug.hs
Loading package ghc-prim ... linking ... done.
Loading package integer-gmp ... linking ... done.
Loading package base ... linking ... done.
Loading package array-0.4.0.1 ... linking ... done.
Loading package deepseq-1.3.0.1 ... linking ... done.
Loading package containers-0.5.0.0 ... linking ... done.
Loading package pretty-1.1.1.0 ... linking ... done.
Loading package template-haskell ... linking ... done.

Bug.hs:7:3:
    Illegal type constructor or class name: `*'
    When splicing a TH declaration:
      foo_0 :: GHC.Types.Bool :: GHC.Prim.*
```

We'd just have to sweeten it back to `StarT` on GHC 7.6.

RyanGlScott:
Fixes #69.

The only true surprise here was that `expand` had to be changed in order to avoid infintely looping when expanding `Type`. See `Note [Don't expand synonyms for *]` for the gory details.

### DIFF
--- a/Language/Haskell/TH/Desugar/Expand.hs
+++ b/Language/Haskell/TH/Desugar/Expand.hs
@@ -4,7 +4,7 @@
 rae@cs.brynmawr.edu
 -}
 
-{-# LANGUAGE CPP, NoMonomorphismRestriction #-}
+{-# LANGUAGE CPP, NoMonomorphismRestriction, ScopedTypeVariables #-}
 
 -----------------------------------------------------------------------------
 -- |
@@ -66,9 +66,9 @@ expand_type ign = go []
       go (t2' : args) t1
     go args (DSigT ty ki) = do
       ty' <- go [] ty
-      return $ foldl DAppT (DSigT ty' ki) args
+      return $ applyDType (DSigT ty' ki) args
     go args (DConT n) = expand_con ign n args
-    go args ty = return $ foldl DAppT ty args
+    go args ty = return $ applyDType ty args
 
 -- | Expands all type synonyms in a desugared predicate.
 expand_pred :: DsMonad q => IgnoreKinds -> DPred -> q DPred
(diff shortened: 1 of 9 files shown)


Language/Haskell/TH/Desugar/Expand.hs
```diff
@@ -168,6 +176,24 @@ expand_con ign n args = do
...
+{-
+Note [Don't expand synonyms for *]
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+We deliberately avoid expanding type synonyms for * such as Type and ★.
+Why? If you reify any such type synonym using Template Haskell, this is
+what you'll get:
+
+  TyConI (TySynD <type synonym name> [] StarT)
+
+If you blindly charge ahead and recursively inspect the right-hand side of
```
goldfirere: Something's very strange here. `Type` and friends are really type synonyms, but they're not recursive. `Type`'s right-hand side should be `TYPE LiftedRep`. See <https://github.com/ghc/ghc/blob/master/libraries/ghc-prim/GHC/Types.hs#L60>. So is TH really saying the RHS of these synonyms is `StarT`? That sounds like a TH bug.

RyanGlScott: > So is TH really saying the RHS of these synonyms is `StarT`?

Yes:

```
$ /opt/ghc/8.4.1/bin/ghci -XTemplateHaskell
GHCi, version 8.4.0.20180118: http://www.haskell.org/ghc/  :? for help
λ> import Language.Haskell.TH hiding (Type)
λ> import Data.Kind
λ> $(reify ''Type >>= stringE . show)
"TyConI (TySynD GHC.Types.Type [] StarT)"
```

> That sounds like a TH bug.

I'm not sure. Look at [these lines](http://git.haskell.org/ghc.git/blob/59fa7b32b018a91f81773ca676251a0b2761ef56:/compiler/deSugar/DsMeta.hs#l994) of `DsMeta`:

```haskell
repTy (HsTyVar _ (L _ n))
  | isLiftedTypeKindTyConName n       = repTStar
```

We do this to avoid cluttering reified Template Haskell kinds with `TYPE 'LiftedRep` all over the place, so a consequence of this is that the right-hand side of `type Type = TYPE 'LiftedRep` will be reified as `StarT`. I suppose we could special-case reifying `Type`/`★`, but is it really worth the hassle? This check works just as well.

goldfirere: I'm afraid I disagree. First off, `isLiftedTypeKindTyConName` should say `False` to `TYPE`. Secondly, I don't think DsMeta is involved at all -- we're reifying here, and so all the code should be in TcSplice. Or am I missing something? The th-desugar check is fine enough, but I still don't see why it's necessary. There are lots of loopy things around in GHC... but I don't think this is one of them.

RyanGlScott: Ack, you're right. I meant [this line of code](http://git.haskell.org/ghc.git/blob/59fa7b32b018a91f81773ca676251a0b2761ef56:/compiler/typecheck/TcSplice.hs#l1710):

```haskell
reifyType ty | isLiftedTypeKind ty = return TH.StarT
```

goldfirere: Ah, well, that's confusing for us in th-desugar, but probably best for the rest of the world. OK. Proceed. :)

PR #74, opened 2018-02-18
Desugar all data types to GADT syntax

### LINKED ISSUE #68: Desugar all datatypes and constructors to GADT syntax
As discussed in #65, the current `th-desugar` AST for datatypes (`DDataD` and `DDataInstD`) cannot express return kinds, such as `data Foo :: forall j. j -> forall k. k -> Type`. It would be natural to add this, but we wish to do so in a way that maintains a minimal surface API. The coexistence of H98 and GADT syntax complicates this story significantly, so we decided in #65 to solve this problem by desugaring _all_ datatypes (and constructors) to GADT syntax, as GADT syntax is a generalization of H98 syntax.

I'm opening this issue because as I attempted to implement this, there were a number of tricky design questions I stumbled upon that I'd be interested in hearing feedback about. Namely:

1. Should we store the return kind of a datatype as `Maybe DKind` or `DKind`? GADT syntax allows for either, but if a return kind is left off, we know that it simply defaults to `Type`, so we could just desugar to `Type` in the absence of an explicit return kind. (Contrast this with `DFamilyResultSig`, which has an explicit `NoSig` form.)
2. How should we store the quantified type variables in each constructor? Currently, `DCon` has a simple `[DTyVarBndr]` field that doesn't state any properties about _what_ type variables it holds. When transitioning everything over to use GADT syntax, however, it becomes less clear what this `[DTyVarBndr]` field does. Does it:

   * Store only the existentially quantified type variables?
   * Store all of the universally quantified type variables in the front of the list, followed by all of the existentially quantified type variables? (Or perhaps we should have two separate lists for the universals and existentials?)

   My inclination is the second option, as there are several spots in `th-desugar` where it's convenient to pluck out the universally quantified type variables (for instance, I need these when sweetening a `DCon` to a `ForallC` on old GHCs).

3. What becomes of the `[DTyVarBndr]` field of `DDataD`/`DDataInstD`? Before, it served double-duty as both the (universal) tyvars of the data type constructor, as well as a syntactic representation of what the user wrote. But if we throw explicit return kinds in the mix, that's no longer true. For instance:

   ```haskell
   data Either (a :: Type) :: Type -> Type where
     Left  :: a -> Either a b
     Right :: b -> Either a b
   ```

   Here, the tyvars of the `Either` tycon are `[a,b]`, but that's _not_ what's shown syntactically (just `[a]`). So the question becomes: should this `[DTyVarBndr]` field show _all_ of the data type constructor's tyvars, or only the ones that were written out syntactically?

### LINKED ISSUE #73: mkExtraTvbs is needlessly complex
While pondering #68, I was reminded of just how horrifically complicated the `mkExtraTvbs` function in `Language.Haskell.TH.Desugar.Core` is:

https://github.com/goldfirere/th-desugar/blob/4ca98c6492015e6ad063d3ad1a2ad6c4f0a56837/Language/Haskell/TH/Desugar/Core.hs#L1004-L1032

This is really, really gross. Why are things this way? It turns out this mess started festering due to a innocent-enough looking test called `dectest9`:

https://github.com/goldfirere/th-desugar/blob/9f4be6dae9a083cccfb1faeeca541afca7a6844d/Test/Splices.hs#L355-L358

Our test suite tries very hard to ensure that when we desugar `Dec9` directly into a `DDec`, it matches what we would get had we reified `Dec9` and then desugared it. In fact, it likely tries _too_ hard.

There's an annoying quirk of GHC that when we reify a datatype with an explicit kind signature like `DDec`, it translates the kind signature into fresh type variables. For example, `Dec9` would likely get reified as:

```haskell
data family Dec9 a (b :: * -> *) (c :: *) -- c is a gensym'd type variable
```

However, this means that we have to do the exact same thing in `th-desugar`, lest the test suite fail. A consequence of this is that we have to _guess the exact same variable names that GHC gensyms_, which is why `mkExtraTvbs` is so complicated in the first place. Ugh.

To make things even worse, @goldfirere discovered that `dectest9` is not robust in the face of recompilation, as rebuilding can sometimes sway GHC to gensym different variable names. Thus, not only is `mkExtraTvbs` overengineered, it doesn't even accomplish the task it was built for!

All of this seems to suggest that our testing should be robust against this sort of behavior. I propose the following change: when testing equality of `DDec`s, we ignore the names of the type variable binders in data types. This isn't as annoying to implement as it sounds, as we're already using a `Data.Data`–based API, so the only change we'd need to the test suite is:

```diff
diff --git a/Test/Splices.hs b/Test/Splices.hs
index 894abb9..c415970 100644
--- a/Test/Splices.hs
+++ b/Test/Splices.hs
@@ -96,10 +96,19 @@ dropTrailing0s = everywhere (mkT (mkName . frob . nameBase))
       | head str == 'R' = str
       | otherwise       = dropWhileEnd isDigit str
 
-eqTH :: (Data a, Show a) => a -> a -> Bool
-eqTH a b = show (unqualify a) == show (unqualify b)
+eqTH :: (Data a, Eq a) => a -> a -> Bool
+eqTH a b = extQ2 (==) eqDDec (unqualify a) (unqualify b)
+  where
+    extQ2 :: (Typeable a, Typeable b)
+          => (a -> a -> q)
+          -> (b -> b -> q)
+          -> a -> a -> q
+    extQ2 f g x y = maybe (f x y) (uncurry g) ((,) <$> cast x <*> cast y)
+
+    eqDDec :: DDec -> DDec -> Bool
+    eqDDec = ...
 
-eqTHSplice :: (Data a, Show a) => a -> a -> Q Exp
+eqTHSplice :: (Data a, Eq a) => a -> a -> Q Exp
 eqTHSplice a b =
   if a `eqTH` b
   then [| True |]
```

(This would require that we derive `Eq` instances for everything in `Language.Haskell.TH.Desugar.Core`, but to be honest, I'm surprised we're not doing this already!)

Once that is done, I think we could replace an enormous chunk of `mkExtraTvbs` with `newName`/`newUniqueName`, which would greatly simplify it.

RyanGlScott:
At long last, this pemits data type declarations in `DDec` to have explicit return kinds without having lots of fiddly `Maybe`s everywhere. Among other benefits, this should make it easier to control CUSK behavior in generated datatypes in `singletons`.

This is a pretty large patch, so here are some highlights:

1. The `DDataD`, `DDataFamilyD`, and `DDataFamInstD` constructors of `DDec` now have `DKind` fields corresponding to explicit return kinds (if desugaring a non-GADT, this defaults to just `Type`). The `DCon` constructor now has a `DType` field instead of `Maybe DType`, since _all_ constructors now have a GADT return type.

2. The `dsCon` function had to be revamped to desugar everything into GADT syntax. Of particular note is when desugaring a _non_-GADT, we have to know what the universally quantified type variables and return type will be, which isn't information that's easily accessible from having just a `Con`. To compensate for this, I've added extra arguments to `dsCon`.

   A consequence of this is that I had to remove the `instance Desugar [Con] [DCon]`, since the `Desugar` interface doesn't give enough information to invoke `dsCon` any more. This is a bit sad, but not entirely unprecedented (after all, there's no `instance Desugar [Pat] [DPat]` instance either, since `dsPat` also has an unorthodox type signature).

3. I added a `conExistentialTvbs` function which has a "dumb-but-predictable" (to steal a phrase of yours, @goldfirere) way to determine the existentially quantified type variables of a constructor. I added lots of documentation to explain the subtleties of this function, reflecting the discussion in https://github.com/goldfirere/th-desugar/issues/68#issuecomment-328913807.

4. Coming up with explicit type variable binders for data family instances is annoying. Like, really annoying—https://github.com/ghc-proposals/ghc-proposals/pull/55 can't come soon enough. To work around this, I have to reverse engineer the type variables in a data family instance and sort them in reverse topological order to emulate what GHC is doing behind the scenes. The `toposortTyVarsOf` function is what accomplishes this.

5. Surprisingly, I didn't have to change the test suite as much as I though! The `Sweeten` module does a darn good job of turning data types back into forms that old GHCs can understand, so even though I did change the  data types in the `dectest` series to use GADT syntax, old GHCs were perfectly capable of handling them.

   There was one small hiccup in porting `imp_inst_test2` and `imp_inst_test3` over to GADT syntax—due to an (apparently undiagnosed) Template Haskell bug on GHC 7.10 and earlier, GHC would simply reject them if written in GADT syntax. I couldn't find a way to work around this, so I simply hacked around the issue with CPP, writing them in non-GADT syntax on older GHCs.

6. Since we now preserve explicit data type return kinds when desugaring, we no longer need to mimic GHC's gensym behavior to make the `dectest`s pass. This means that `mkExtraDKindBinders` (what used to be called `mkExtraTvbs`) can now be drastically simplified to just use `newName`. Thus, this patch fixes #73.

Along the way, I also added some functions which I needed when porting `singletons` over to accommodate these changes. I'll post the accompanying `singletons` patch shortly. **Edit**: At https://github.com/goldfirere/singletons/pull/305.

Fixes #68.

### DIFF
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -1614,8 +1614,7 @@ fvDType :: DType -> S.Set Name
 fvDType = go_ty
   where
     go_ty :: DType -> S.Set Name
-    go_ty (DForallT tvbs cxt ty) = (foldMap go_tvb tvbs <> go_ty ty <> foldMap go_pred cxt)
-                                   S.\\ S.fromList (map dtvbName tvbs)
+    go_ty (DForallT tvbs cxt ty) = foldr go_tvb (foldMap go_pred cxt <> go_ty ty) tvbs
     go_ty (DAppT t1 t2)          = go_ty t1 <> go_ty t2
     go_ty (DSigT ty ki)          = go_ty ty <> go_ty ki
     go_ty (DVarT n)              = S.singleton n
@@ -1630,9 +1629,9 @@ fvDType = go_ty
     go_pred (DVarPr n)     = S.singleton n
     go_pred _              = S.empty
 
-    go_tvb :: DTyVarBndr -> S.Set Name
-    go_tvb (DPlainTV{})    = S.empty
-    go_tvb (DKindedTV _ k) = go_ty k
+    go_tvb :: DTyVarBndr -> S.Set Name -> S.Set Name
+    go_tvb (DPlainTV n)    fvs = S.delete n fvs
+    go_tvb (DKindedTV n k) fvs = S.delete n fvs <> go_ty k
 
 dtvbName :: DTyVarBndr -> Name
 dtvbName (DPlainTV n)    = n
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -25,6 +25,10 @@ import Control.Monad hiding (forM_, mapM)
 import Control.Monad.Zip
 import Control.Monad.Writer hiding (forM_, mapM)
 import Data.Foldable hiding (notElem)
+import Data.Graph
+import qualified Data.Map as Map
+import Data.Map (Map)
+import qualified Data.Set as Set
 import Data.Traversable
 import Data.Data hiding (Fixity)
 #if __GLASGOW_HASKELL__ > 710
@@ -120,15 +124,15 @@ data NewOrData = Newtype
 
 -- | Corresponds to TH's @Dec@ type.
 data DDec = DLetDec DLetDec
-          | DDataD NewOrData DCxt Name [DTyVarBndr] [DCon] [DDerivClause]
+          | DDataD NewOrData DCxt Name [DTyVarBndr] DKind [DCon] [DDerivClause]
           | DTySynD Name [DTyVarBndr] DType
           | DClassD DCxt Name [DTyVarBndr] [FunDep] [DDec]
           | DInstanceD (Maybe Overlap) DCxt DType [DDec]
           | DForeignD DForeign
           | DOpenTypeFamilyD DTypeFamilyHead
           | DClosedTypeFamilyD DTypeFamilyHead [DTySynEqn]
-          | DDataFamilyD Name [DTyVarBndr]
-          | DDataInstD NewOrData DCxt Name [DType] [DCon] [DDerivClause]
+          | DDataFamilyD Name [DTyVarBndr] DKind
+          | DDataInstD NewOrData DCxt Name [DType] DKind [DCon] [DDerivClause]
           | DTySynInstD Name DTySynEqn
           | DRoleAnnotD Name [Role]
           | DStandaloneDerivD (Maybe DerivStrategy) DCxt DType
@@ -176,9 +180,23 @@ data InjectivityAnn = InjectivityAnn Name [Name]
   deriving (Eq, Ord, Show, Typeable, Data, Generic)
 #endif
 
--- | Corresponds to TH's @Con@ type.
+-- | Corresponds to TH's 'Con' type. Unlike 'Con', all 'DCon's reflect GADT
+-- syntax. This is beneficial for @th-desugar@'s since it means
+-- that all data type declarations can support explicit return kinds, so
+-- one does not need to represent them with something like @'Maybe' 'DKind'@,
+-- since Haskell98-style data declaration syntax isn't used. Accordingly,
+-- there are some differences between 'DCon' and 'Con' to keep in mind:
+--
+-- * Unlike 'ForallC', where the meaning of the 'TyVarBndr's changes depending
+--   on whether it's followed by 'GadtC'/'RecGadtC' or not, the meaning of the
+--   'DTyVarBndr's in a 'DCon' is always the same: it is the list of
+--   universally /and/ existentially quantified type variables. Note that it is
+--   not guaranteed that one set of type variables will appear before the
+--   other.
+--
+-- * A 'DCon' always has an explicit return type.
 data DCon = DCon [DTyVarBndr] DCxt Name DConFields
-                 (Maybe DType)  -- ^ A GADT result type, if there is one
+                 DType  -- ^ The GADT result type
           deriving (Show, Typeable, Data, Generic)
 
 -- | A list of fields either for a standard data constructor or a record
@@ -849,7 +867,7 @@ fixBug8884ForFamilies (DClosedTypeFamilyD (DTypeFamilyHead name tvbs frs ann) eq
       eqns' = map (fixBug8884ForEqn num_args) eqns
   frs' <- remove_arrows num_args frs
   return (DClosedTypeFamilyD (DTypeFamilyHead name tvbs frs' ann) eqns', num_args)
-fixBug8884ForFamilies dec@(DDataFamilyD _ _)
+fixBug8884ForFamilies dec@(DDataFamilyD _ _ _)
   = return (dec, 0)   -- the num_args is ignored for data families
 fixBug8884ForFamilies dec =
   impossible $ "Reifying yielded a FamilyI with a non-family Dec: " ++ show dec
@@ -894,25 +912,33 @@ dsDec d@(FunD {}) = (fmap . map) DLetDec $ dsLetDec d
 dsDec d@(ValD {}) = (fmap . map) DLetDec $ dsLetDec d
 #if __GLASGOW_HASKELL__ > 710
 dsDec (DataD cxt n tvbs mk cons derivings) = do
-  extra_tvbs <- mkExtraTvbs tvbs mk
+  tvbs' <- mapM dsTvb tvbs
+  let data_type = vanillaDataReturnType n tvbs'
   (:[]) <$> (DDataD Data <$> dsCxt cxt <*> pure n
-                         <*> ((++ extra_tvbs) <$> mapM dsTvb tvbs)
-                         <*> concatMapM dsCon cons
+                         <*> mapM dsTvb tvbs <*> maybeDsReturnKind mk
+                         <*> concatMapM (dsCon tvbs' data_type) cons
                          <*> mapM dsDerivClause derivings)
 dsDec (NewtypeD cxt n tvbs mk con derivings) = do
-  extra_tvbs <- mkExtraTvbs tvbs mk
+  tvbs' <- mapM dsTvb tvbs
+  let data_type = vanillaDataReturnType n tvbs'
   (:[]) <$> (DDataD Newtype <$> dsCxt cxt <*> pure n
-                            <*> ((++ extra_tvbs) <$> mapM dsTvb tvbs)
-                            <*> dsCon con
+                            <*> mapM dsTvb tvbs <*> maybeDsReturnKind mk
+                            <*> dsCon tvbs' data_type con
                             <*> mapM dsDerivClause derivings)
 #else
-dsDec (DataD cxt n tvbs cons derivings) =
+dsDec (DataD cxt n tvbs cons derivings) = do
+  tvbs' <- mapM dsTvb tvbs
+  let data_type = vanillaDataReturnType n tvbs'
   (:[]) <$> (DDataD Data <$> dsCxt cxt <*> pure n
-                         <*> mapM dsTvb tvbs <*> concatMapM dsCon cons
+                         <*> pure tvbs' <*> pure typeKind
+                         <*> concatMapM (dsCon tvbs' data_type) cons
                          <*> mapM dsDerivClause derivings)
-dsDec (NewtypeD cxt n tvbs con derivings) =
+dsDec (NewtypeD cxt n tvbs con derivings) = do
+  tvbs' <- mapM dsTvb tvbs
+  let data_type = vanillaDataReturnType n tvbs'
   (:[]) <$> (DDataD Newtype <$> dsCxt cxt <*> pure n
-                            <*> mapM dsTvb tvbs <*> dsCon con
+                            <*> pure tvbs' <*> pure typeKind
+                            <*> dsCon tvbs' data_type con
                             <*> mapM dsDerivClause derivings)
 #endif
 dsDec (TySynD n tvbs ty) =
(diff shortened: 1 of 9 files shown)


RyanGlScott:
Ping?

goldfirere:
Sorry, Ryan. I'm very swamped until the ICFP deadline. This is a biggish patch. If you like, you can merge (I still think I'll be able to review it), or I can get to it in a few weeks.

RyanGlScott:
Alright, do you have a chance to review it now? :)

(I definitely don't want to merge this without another pair of eyeballs looking over it—there's lots of tricky stuff going on here.)

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1437,3 +1505,141 @@ dTypeToDPred (DConT n)       = return $ DConPr n
...
+
+typeKind :: DKind
+typeKind = DConT typeKindName
+
+maybeDsReturnKind :: DsMonad q => Maybe Kind -> q DKind
+maybeDsReturnKind = maybe (return typeKind) dsType
+
+-- Take a vanilla (i.e., non–data-family-instance) data type name and
+-- apply it to its type variable binders to form a DType.
+vanillaDataReturnType :: Name -> [DTyVarBndr] -> DType
```
RyanGlScott: It just occurred to me that the use of "vanilla" here is inconsistent with how it's used in https://github.com/goldfirere/singletons/pull/305/files#diff-b095422bfab81c31d45ee204e13c4900R284. @goldfirere, can you think of a better name for this function?

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -894,25 +912,33 @@ dsDec d@(FunD {}) = (fmap . map) DLetDec $ dsLetDec d
 dsDec d@(ValD {}) = (fmap . map) DLetDec $ dsLetDec d
 #if __GLASGOW_HASKELL__ > 710
 dsDec (DataD cxt n tvbs mk cons derivings) = do
-  extra_tvbs <- mkExtraTvbs tvbs mk
+  tvbs' <- mapM dsTvb tvbs
+  let data_type = vanillaDataReturnType n tvbs'
```
goldfirere: `vanillaDataReturnType` will do the wrong thing if the `tvbs` list is incomplete (like in `data T :: Type -> Type where ...`, with an empty `tvbs`). It turns out that this doesn't matter, because `data_type` is ignored for GADT-syntax declarations, which are the only ones that might have a foreshortened `tvbs` list. So I don't spot a bug here, but I think some comments are in order explaining that `data_type` is valid only in the H98 case.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -894,25 +912,33 @@ dsDec d@(FunD {}) = (fmap . map) DLetDec $ dsLetDec d
...
   (:[]) <$> (DDataD Data <$> dsCxt cxt <*> pure n
-                         <*> ((++ extra_tvbs) <$> mapM dsTvb tvbs)
-                         <*> concatMapM dsCon cons
+                         <*> mapM dsTvb tvbs <*> maybeDsReturnKind mk
```
goldfirere: Is there a reason to use `mapM dsTvb tvbs` here instead of `tvbs'`? It doesn't look like it.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -894,25 +912,33 @@ dsDec d@(FunD {}) = (fmap . map) DLetDec $ dsLetDec d
...
+                         <*> concatMapM (dsCon tvbs' data_type) cons
                          <*> mapM dsDerivClause derivings)
 dsDec (NewtypeD cxt n tvbs mk con derivings) = do
-  extra_tvbs <- mkExtraTvbs tvbs mk
+  tvbs' <- mapM dsTvb tvbs
+  let data_type = vanillaDataReturnType n tvbs'
   (:[]) <$> (DDataD Newtype <$> dsCxt cxt <*> pure n
-                            <*> ((++ extra_tvbs) <$> mapM dsTvb tvbs)
-                            <*> dsCon con
+                            <*> mapM dsTvb tvbs <*> maybeDsReturnKind mk
```
goldfirere: Another repeated `mapM dsTvb tvbs`.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -934,37 +960,48 @@ dsDec d@(PragmaD {}) = (fmap . map) DLetDec $ dsLetDec d
...
-dsDec (FamilyD DataFam n tvbs m_k) = do
-  extra_tvbs <- mkExtraTvbs tvbs m_k
-  (:[]) <$> (DDataFamilyD n <$> ((++ extra_tvbs) <$> mapM dsTvb tvbs))
+dsDec (FamilyD DataFam n tvbs m_k) =
+  (:[]) <$> (DDataFamilyD n <$> mapM dsTvb tvbs <*> maybeDsReturnKind m_k)
 #endif
 #if __GLASGOW_HASKELL__ > 710
 dsDec (DataInstD cxt n tys mk cons derivings) = do
-  extra_tvbs <- map dTyVarBndrToDType <$> mkExtraTvbs [] mk
+  tys' <- dataFamInstTypes tys mk
```
goldfirere: There is repeated work here. `dataFamInstTypes` desugars the types, and then this is done again by the `mapM dsType tys` below. Instead, perhaps inline `dataFamInstTypes` here and then you don't have to repeat the work. (I say "inline" because once you desugar the types here, `dataFamInstTypes` is little more than a call to `mkExtraKindBinders`.)

Similarly, it seems there is wasted work in `maybeDsReturnKind` below.

Language/Haskell/TH/Desugar/Core.hs
goldfirere: Agreed that "vanilla" is bad here. I would just say non-family.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1437,3 +1505,141 @@ dTypeToDPred (DConT n)       = return $ DConPr n
...
+-- variable names.
+dataFamInstTypes :: DsMonad q => [Type] -> Maybe Kind -> q [DType]
+dataFamInstTypes tys mk = do
+  tys' <- mapM dsType tys
+  extra_tvbs <- mkExtraKindBinders mk
+  return $ tys' ++ map dTyVarBndrToDType extra_tvbs
+
+-- Unlike vanilla data types and data family declarations, data family
+-- instance declarations do not come equipped with a list of bound type
+-- variables (at least not yet—see Trace #14268). This means that we have
```
goldfirere: s/Trace/Trac/

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1437,3 +1505,141 @@ dTypeToDPred (DConT n)       = return $ DConPr n
...
+-- returned list is well scoped.
+dataFamInstTvbs :: [DType] -> [DTyVarBndr]
+dataFamInstTvbs = toposortTyVarsOf
+
+-- | Take a list of 'DType's, find their free variables, and sort them in
+-- reverse topological order to ensure that they are well scoped.
+--
+-- On older GHCs, this takes measures to avoid returning explicitly bound
+-- kind variables, which was not possible before @TypeInType@.
+toposortTyVarsOf :: [DType] -> [DTyVarBndr]
```
goldfirere: Why is this necessary? It smells a bit like type inference, especially with the explicit search for kind signatures. I suppose that without this search, we'd be unable to order the tyvars correctly in data constructors. And perhaps we want to promise to our users that data constructors list all their variables. OK. You've convinced me. :) But I don't like it.

It will all be better once https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0007-instance-foralls.rst is implemented, and then users have a better way of specifying variable kinds. (We would not do this kind-sig search, sometimes failing if we get it wrong, but then users would have easy recourse.) Argh. I suppose even then we'd have to keep this around to support a variety of GHC versions.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1437,3 +1505,141 @@ dTypeToDPred (DConT n)       = return $ DConPr n
...
+
+  in map ascribeWithKind $
+     filter (not . isKindBinderOnOldGHCs) $
+     map lookupVertex tg
+
+fvDType :: DType -> S.Set Name
+fvDType = go_ty
+  where
+    go_ty :: DType -> S.Set Name
+    go_ty (DForallT tvbs cxt ty) = (foldMap go_tvb tvbs <> go_ty ty <> foldMap go_pred cxt)
```
goldfirere: This is wrong in the presence of dependencies within the `tvbs`. You'll need to take them one-at-a-time.

Language/Haskell/TH/Desugar/Sweeten.hs
```diff
@@ -217,34 +217,67 @@ letDecToTH (DPragmaD prag)      = fmap PragmaD (pragmaToTH prag)
...
 conToTH (DCon tvbs cxt n fields rty) =
   ForallC (map tvbToTH tvbs) (cxtToTH cxt) (conToTH $ DCon [] [] n fields rty)
+#else
+-- On GHCs earlier than 8.0, we must be careful, since the only time ForallC is
+-- used is when there are either:
+--
+-- 1. Any existentially quantified type variables
+-- 2. A constructor context
+--
+-- If neither of these conditions hold, then we must avoid the use of ForallC.
```
goldfirere: Why? What goes wrong if we use `ForallC`?

Language/Haskell/TH/Desugar/Core.hs
goldfirere: Suppose we have `forall k. (forall (a :: k) (k :: Type) (b :: k). Proxy b -> Proxy a) -> Proxy k` and we're asking what the free variables are of the first argument of this function. Note that `k` is shadowed. The current implementation will conclude that there are no free variables, even though the outer `k` really is free there (in the kind of `a`).

I'm sure you're right that the current implementation is wrong! Not too hard to fix, though, with a `mapAccumL` or some such.

RyanGlScott: I've just pushed a commit which (I believe) implements this suggestion.

RyanGlScott: I've also added your counterexample as a test case.

goldfirere:
Ship it!

PR #82, opened 2018-06-17
Desugar partialilty in `do`-notation to `fail`

### LINKED ISSUE #80: th-desugar doesn't desugar partial pattern-matches in do-notation to `fail`
Take this function:

```haskell
f :: Maybe Int
f = do
  Just x <- Nothing
  return x
```

Despite its appearance, `f` is a total function, since the apparently partial pattern-match desugars to something resembling:

```haskell
case Nothing of
  Just x -> return x
  Nothing -> fail "Pattern match failure in do expression"
```

However, `th-desugar` doesn't respect this property during its own desugaring, as demonstrated in this GHCi session:

```
λ> putStrLn $([d| f :: Maybe Int; f = do { Just x <- Nothing; return x } |] >>= dsDecs >>= stringE . groom)
[DLetDec
   (DSigD f_6989586621679077736
      (DAppT (DConT GHC.Base.Maybe) (DConT GHC.Types.Int))),
 DLetDec
   (DValD (DVarPa f_6989586621679077736)
      (DAppE (DAppE (DVarE GHC.Base.>>=) (DConE GHC.Base.Nothing))
         (DLamE [arg_6989586621679077738_6989586621679077739]
            (DCaseE (DVarE arg_6989586621679077738_6989586621679077739)
               [DMatch (DConPa GHC.Base.Just [DVarPa x_6989586621679077737])
                  (DAppE (DVarE GHC.Base.return) (DVarE x_6989586621679077737))]))))]
```

If you squint at that long enough, you'll eventually realize that it's desugaring `Just x <- Nothing` to `case Nothing of { Just x -> return x }`, which is less defined than it should be (after all, the overridden `fail` implementation in the `Monad []` instance is what makes this trick work). We should implement the `fail` desugaring in `th-desugar` as well to compensate for this.

RyanGlScott:
Fixes #80.

### DIFF
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -7,7 +7,7 @@ Desugars full Template Haskell syntax into a smaller core syntax for further
 processing. The desugared types and constructors are prefixed with a D.
 -}
 
-{-# LANGUAGE TemplateHaskell, LambdaCase, CPP, TupleSections #-}
+{-# LANGUAGE TemplateHaskell, LambdaCase, CPP, ScopedTypeVariables, TupleSections #-}
 
 module Language.Haskell.TH.Desugar.Core where
 
@@ -31,6 +31,10 @@ import qualified Data.Set as Set
 import Data.Traversable
 #if __GLASGOW_HASKELL__ > 710
 import Data.Maybe (isJust)
+#endif
+
+#if __GLASGOW_HASKELL__ >= 800
+import qualified Control.Monad.Fail as MonadFail
 #endif
 
 #if __GLASGOW_HASKELL__ >= 803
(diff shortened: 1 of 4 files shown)


Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -669,6 +667,40 @@ dsComp (ParS stmtss : rest) = do
...
   DAppE (DAppE (DVarE '(>>=)) exp) <$> dsLam [pat] rest'
 
+-- Desugar a binding statement in a do- or list comprehension.
+--
+-- In the event that the pattern in the statement is partial, the desugared
+-- case expression will contain a catch-all case that calls 'fail' from either
+-- 'MonadFail' or 'Monad', depending on whether the @MonadFailDesugaring@
+-- language extension is enabled or not. (On GHCs older than 8.0, 'fail' from
+-- 'Monad' is always used.)
+dsBindS :: forall q. DsMonad q => Exp -> Pat -> q DExp -> String -> q DExp
```
goldfirere: Can we change the `q DExp` param to be just a `DExp`? Having it in the `q` monad makes me think that the function changes the context somehow before running that action.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -669,6 +667,40 @@ dsComp (ParS stmtss : rest) = do
...
+             fail_name <- mk_fail_name
+             return $ bind_into $ DLamE [arg_name] $ DCaseE (DVarE arg_name)
+               [ DMatch success_pat' success_exp'
+               , DMatch DWildPa $
+                 DVarE fail_name `DAppE`
+                   DLitE (StringL $ "Pattern match failure in " ++ ctxt)
+               ]
+  where
+    mk_fail_name :: q Name
+#if __GLASGOW_HASKELL__ >= 800
```
goldfirere: I thought you said that `qIsExtEnabled` is from 8.2 and up...

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: Sure. I only did this to save some boilerplate at `dsBindS`'s call sites, but I can see how this would be confusing.

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: I misremembered that (see https://github.com/goldfirere/th-desugar/issues/80#issuecomment-397843906).

PR #94, opened 2018-10-22
Fix #92 by filling in toposortTyVarsOf's missing case

### LINKED ISSUE #92: toposortTyVarsOf doesn't handle foralls
When I originally wrote the code for `toposortTyVarsOf`, I left a bit of partiality in:

https://github.com/goldfirere/th-desugar/blob/1b76cf6da5a28ad133126aa77201080ff0d83f51/Language/Haskell/TH/Desugar/Core.hs#L1326-L1330

At the time, I thought that this was OK, since `toposortTyVarsOf` would only ever be called on type family patterns, which shouldn't ever contain `forall`s. But I later found _other_ uses for `toposortTyVarsOf`, such as in `singletons`, where I use it to properly sort the free variables in defunctionalization symbols' return kinds. And wouldn't you know it, those variables _can_ have polytypes appear in their kinds. As an example of this, the following piece of `singletons` code crashes:

```haskell
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE RankNTypes #-}
module Bug where

import Data.Singletons.TH

$(singletons [d|
  type Foo (x :: y) (f :: forall a. a) = f Bool
  |])
```
```
$ /opt/ghc/8.6.1/bin/ghc Bug.hs

Bug.hs:1:1: error:
    Exception when trying to run compile-time code:
      `forall` type used in type family pattern
CallStack (from HasCallStack):
  error, called at ./Language/Haskell/TH/Desugar/Core.hs:1330:30 in th-desugar-1.9-IWJeiZlphUgItyl5W2YZ4A:Language.Haskell.TH.Desugar.Core
    Code: singletons
            [d| type Foo (x :: y) (f :: forall a. a) = f Bool |]
  |
1 | {-# LANGUAGE PolyKinds #-}
  | ^
```

(Not that I expect that this code _should_ typecheck necessarily, since trying to defunctionalize a type synonym with a higher-ranked argument like this has other problems. But at the very least, it shouldn't crash.)

In light of this, it seems that we should just bite the bullet and implement the `DForallT` case of `toposortTyVarsOf` properly. Here is a minimal test case which triggers the `error`:

```haskell
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE RankNTypes #-}
module Bug where

import Language.Haskell.TH

$(do let a = mkName "a"
     t  <- [t| forall f. f $(varT a) |]
     t' <- desugar t
     runIO $ print $ toposortTyVarsOf [t']
     pure [])
```

RyanGlScott:
That'll teach me to leave `error`s in my code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 `th-desugar` release notes
 ==========================
 
+Version next
+------------
+* Fix a bug in which `toposortTyVarsOf` would error at runtime if given types
+  containing `forall`s as arguments.
+
 Version 1.9
 -----------
 * Suppose GHC 8.6.
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -1324,21 +1324,38 @@ toposortTyVarsOf tys =
       fvs = Set.toList $ foldMap fvDType tys
 
       varKindSigs :: Map Name DKind
-      varKindSigs = foldMap go tys
+      varKindSigs = foldMap go_ty tys
         where
-          go :: DType -> Map Name DKind
-          go (DForallT {}) = error "`forall` type used in type family pattern"
-          go (DAppT t1 t2) = go t1 `mappend` go t2
-          go (DSigT t k) =
-            let kSigs = go k
+          go_ty :: DType -> Map Name DKind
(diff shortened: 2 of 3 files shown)


goldfirere:
LGTM

PR #95, opened 2018-10-22
Introduce Language.Haskell.TH.Desugar.FV

### LINKED ISSUE #93: fvDType returns incorrect results for quantified constraints
Consider the following program:

```haskell
{-# LANGUAGE QuantifiedConstraints #-}
module Bug where

import Language.Haskell.TH

$(do let a = mkName "a"
     t  <- [t| (Show $(varT a) => Show (Maybe $(varT a))) => String |]
           -- (Show a => Show (Maybe a)) => String
     t' <- desugar t
     runIO $ print $ fvDType t'
     pure [])
```

If you run this, it'll claim that the free variables of that type (`(Show a => Show (Maybe a)) => String`) are:

```
$ /opt/ghc/8.6.1/bin/ghc Bug.hs
fromList []
```

This is completely wrong, since `a` is free in that type! The reason this happens is because `fvDType` is missing a case for `DForallPr`.

I've fixed this on the `FVs` branch as part of a larger effort, but this would be worth fixing on its own.

RyanGlScott:
This migrates all of the FV-computing functions to their own module. It also adds some new FV-computing functions (e.g., `fvDExp`) and fixes bugs in existing ones (e.g., this fixes #93).


goldfirere:
I didn't read carefully, but the general idea looks good.

PR #104, opened 2018-12-29
Fix #102 and #103

### LINKED ISSUE #102: getRecordSelectors is horribly broken for GADTs
Consider this example:

```haskell
#!/usr/bin/env cabal
{- cabal:
build-depends: base, template-haskell, th-desugar
-}
{-# LANGUAGE GADTs #-}
{-# LANGUAGE ScopedTypeVariables #-}
module Main (main) where

import Language.Haskell.TH

main :: IO ()
main = putStrLn
  $(do [DDataD _ _ foo1 [DPlainTV x1] _ cons1 _] <-
         [d| data Foo1 x = MkFoo1 { unFoo1 :: x } |] >>= dsDecs
       recs1 <- getRecordSelectors (DConT foo1 `DAppT` DVarT x1) cons1

       [DDataD _ _ foo2 [DPlainTV x2] _ cons2 _] <-
         [d| data Foo2 x where
              MkFoo2 :: forall a. { unFoo2 :: a } -> Foo2 a |] >>= dsDecs
       recs2 <- getRecordSelectors (DConT foo2 `DAppT` DVarT x2) cons2
       stringE $ unlines [ "Foo1:"
                         , show recs1
                         , "Foo2:"
                         , show recs2
                         ])
```

Here, we call `getRecordSelectors` on two equivalent data types, `Foo1` and `Foo2`, where one is written as a normal ADT, and one is written with GADT syntax. Despite being equivalent, `getRecordSelectors` returns different results for them:

```
$ cabal new-run Bug.hs
Resolving dependencies...
Build profile: -w ghc-8.6.3 -O1

<elided>

Foo1:
[DSigD unFoo1_6989586621679043062 (DForallT [DPlainTV x_6989586621679043063] [] (DAppT (DAppT DArrowT (DAppT (DConT Foo1_6989586621679043060) (DVarT x_6989586621679043063))) (DVarT x_6989586621679043063))),DFunD unFoo1_6989586621679043062 [DClause [DConPa MkFoo1_6989586621679043061 [DVarPa field_6989586621679043064]] (DVarE field_6989586621679043064)]]
Foo2:
[]
```

The `unFoo1` record selector is correctly extracted, but `unFoo2` is not. The issue that `getRecordSelectors` assumes that the universally quantified type variables in a data constructor are the exact same as the ones bound by the data type head. This assumption happens to be true for normal ADTs, but not GADTs.

### LINKED ISSUE #103: dsCon doesn't quantify kind variables
```haskell
#!/usr/bin/env cabal
{- cabal:
build-depends: base, template-haskell, th-desugar
-}
{-# LANGUAGE PolyKinds #-}
module Main where

import Language.Haskell.TH

main :: IO ()
main = putStrLn
 $(do decs <- [d| data P (a :: k) = MkP |]
      decs' <- dsDecs decs
      stringE $ show decs')
```

This produces:

```
[DDataD Data [] P_6989586621679036346 [DKindedTV a_6989586621679036349 (DVarT k_6989586621679036348)] Nothing [DCon [DKindedTV a_6989586621679036349 (DVarT k_6989586621679036348)] [] MkP_6989586621679036347 (DNormalC False []) (DAppT (DConT P_6989586621679036346) (DSigT (DVarT a_6989586621679036349) (DVarT k_6989586621679036348)))] []]
```

Or, cleaned up:

```haskell
data P (a :: k) where
  MkP :: forall (a :: k). P (a :: k)
```

The type of `MkP` really ought to be `forall k (a :: k). P (a :: k)`, however. Patch incoming.

RyanGlScott:
A two-for-one special.

This fixes #102 by tweaking `getRecordSelectors` to actually use `conExistentialTvbs`, instead of the horribly ad hoc tricks it was using before to detect which type variables are existentially quantified.

This fixes #103 with a simple use of `toposortTyVarsOf`.

### DIFF
--- a/Language/Haskell/TH/Desugar.hs
+++ b/Language/Haskell/TH/Desugar.hs
@@ -111,6 +111,8 @@ import Language.Haskell.TH.Desugar.Match
 import Language.Haskell.TH.Desugar.Subst
 
 import Control.Monad
+import Data.Function
+import Data.List
 import qualified Data.Map as M
 import qualified Data.Set as S
 import Prelude hiding ( exp )
@@ -207,28 +209,32 @@ flattenDValD other_dec = return [other_dec]
 -- See https://github.com/goldfirere/singletons/issues/180 for an example where
 -- the latter behavior can bite you.
 
-getRecordSelectors :: Quasi q
+getRecordSelectors :: DsMonad q
                    => DType        -- ^ the type of the argument
                    -> [DCon]
                    -> q [DLetDec]
 getRecordSelectors arg_ty cons = merge_let_decs `fmap` concatMapM get_record_sels cons
   where
-    get_record_sels (DCon _ _ con_name con _) = case con of
-      DRecC fields -> go fields
-      _ -> return []
-      where
-        go fields = do
-          varName <- qNewName "field"
-          let tvbs     = fvDType arg_ty
-              forall'  = DForallT (map DPlainTV $ S.toList tvbs) []
-              num_pats = length fields
-          return $ concat
-            [ [ DSigD name (forall' $ DArrowT `DAppT` arg_ty `DAppT` res_ty)
-              , DFunD name [DClause [DConPa con_name (mk_field_pats n num_pats varName)]
-                                    (DVarE varName)] ]
-            | ((name, _strict, res_ty), n) <- zip fields [0..]
-            , fvDType res_ty `S.isSubsetOf` tvbs   -- exclude "naughty" selectors
-            ]
+    get_record_sels con@(DCon con_tvbs _ con_name con_fields con_ret_ty) =
+      case con_fields of
+        DRecC fields -> go fields
+        DNormalC{}   -> return []
+        where
+          go fields = do
+            varName <- qNewName "field"
+            con_ex_tvbs <- conExistentialTvbs arg_ty con
+            let con_univ_tvbs  = deleteFirstsBy ((==) `on` dtvbName) con_tvbs con_ex_tvbs
+                con_ex_tvb_set = S.fromList $ map dtvbName con_ex_tvbs
+                forall'        = DForallT con_univ_tvbs []
+                num_pats       = length fields
+            return $ concat
+              [ [ DSigD name (forall' $ DArrowT `DAppT` con_ret_ty `DAppT` field_ty)
+                , DFunD name [DClause [DConPa con_name (mk_field_pats n num_pats varName)]
+                                      (DVarE varName)] ]
+              | ((name, _strict, field_ty), n) <- zip fields [0..]
+              , S.null (fvDType field_ty `S.intersection` con_ex_tvb_set)
+                  -- exclude "naughty" selectors
+              ]
 
     mk_field_pats :: Int -> Int -> Name -> [DPat]
     mk_field_pats 0 total name = DVarPa name : (replicate (total-1) DWildPa)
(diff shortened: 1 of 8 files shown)


Language/Haskell/TH/Desugar.hs
```diff
@@ -369,23 +375,16 @@ conExistentialTvbs :: DsMonad q
                    => DType -- ^ The type of the original data declaration
                    -> DCon
                    -> q [DTyVarBndr]
-conExistentialTvbs data_ty (DCon tvbs _ _ _ ret_ty) =
-  -- Due to GHC Trac #13885, it's possible that the type variables bound by
-  -- a GADT constructor will shadow those that are bound by the data type.
-  -- This function assumes this isn't the case in certain parts (e.g., when
-  -- unifying types), so we do an alpha-renaming of the
-  -- constructor-bound variables before proceeding.
-  substTyVarBndrs M.empty tvbs $ \subst tvbs' -> do
```
RyanGlScott: In retrospect, this use of `substTyVarBndrs` doesn't really make sense. I think I rationalized its use here since I cargo-culted this code from `th-abstraction`, where it is absolutely vital that constructor-bound type variables do not shadow data type-bound ones. That requirement is not present here, and moreover, this actually freshens type variables _too_ much, since the variables that this function returns are different from the ones that the constructor binds! It's incredible how long this buglet has went unnoticed.

Thankfully, it appears that we can just remove this use of `substTyVarBndrs` entirely, as it's unneeded. (I verified that `singletons` still works with these changes as well.)

RyanGlScott: > What about clients of th-desugar that use a GHC that hasn't fixed GHC#13885?

Those will still work. Again, GHC#13885 really has nothing to do with this—that was a leftover from [this code from `th-abstraction`](https://github.com/glguy/th-abstraction/blob/14d02a4e479ab37dd2464553a83bc8df9b7df89b/src/Language/Haskell/TH/Datatype.hs#L892-L910), on which I adapted this code in `th-desugar`. Since it seemed to work in `th-desugar` without much fuss, I didn't really stop to think about this comment much, but I now realize that what we're doing in `th-abstraction` is subtly different from what we're doing in `th-desugar`, and that the invariants that must hold in `th-abstraction` do not apply to `th-desugar`.

In `th-abstraction`, we not only identify existentially quantified variables, but also existential _constraints_, including equality constraints. It's the latter thing that requires GHC#13885 to be fixed, since having duplicate occurrences of variables could lead to extraneous equality constraints like `a ~ a`. In `th-desugar`, this isn't a concern, so we don't need to perform this freshening step (and in fact this bites us, as I explain in my other reply below).

As evidence for my claim that this actually works, there's a test case that stress-tests this: `Test.DsDec`, which ensures that `getRecordSelectors` (which now uses `conExistentialTvbs`) detects "naughty" record selectors (i.e., those with existential type variables in their types). This works despite the fact that this test case works on a normal ADT, and when we desugar normal ADTs, we reuse that same type variables from the data type head in each constructor, which means that we're effectively under the same conditions as in GHC#13885.

> I don't understand the "freshen too much" part, because the substitution is applied to the return type in the constructor, which is duly unified with the data type.

This one is simple to explain. If you have this constructor:

```haskell
data Foo where
  MkFoo :: a_0 -> Foo
```

Since `substTyVarBndrs` always freshens bound variables, we would end up returning `[a_1]`, not `[a_0]`, it we invoke `conExistentialTvbs` on `MkFoo`! This defeats the whole purpose of `conExistentialTvbs`, and more importantly, it causes issues when used in `getRecordSelectors`, since we have to subtract out the existential type variables and it's vital that we extract the right ones, uniques and all.

goldfirere:
I took only a cursory glance through, but it looks good, modulo yet another use of `toposortTyVars`, which I claim is bogus. See https://github.com/goldfirere/th-desugar/pull/105#discussion_r244869756

Issue #88, opened 2018-08-28
Locally reified class methods don't explicitly quantify kind variables


RyanGlScott:
If you run the following program:

```haskell
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TemplateHaskell #-}
module Main where

import Language.Haskell.TH
import Language.Haskell.TH.Desugar

main :: IO ()
main = putStrLn
 $([d| class C (f :: k -> Type) where
         method :: forall (a :: k). f a
     |] >>= \ds -> withLocalDeclarations ds (dsReify (mkName "method")) >>= stringE . show)
```

You'll get the following type for `method`:

```
$ /opt/ghc/8.4.3/bin/runghc Bug2.hs
Just (DVarI method (DForallT [DKindedTV f_6989586621679035180 (DAppT (DAppT DArrowT (DVarT k_6989586621679035179)) (DConT Language.Haskell.TH.Syntax.Type))] [DAppPr (DConPr C_6989586621679035177) (DVarT f_6989586621679035180)] (DForallT [DKindedTV a_6989586621679035181 (DVarT k_6989586621679035179)] [] (DAppT (DVarT f_6989586621679035180) (DVarT a_6989586621679035181)))) (Just C_6989586621679035177))
```

After some cleanup, that's:

```haskell
method :: forall (f :: k -> Type). C f => forall (a :: k). f a
```

This type isn't wrong today, but once https://github.com/ghc-proposals/ghc-proposals/pull/103 is implemented, this will be incorrect, since `k` is implicitly brought into scope in a top-level `forall`.

RyanGlScott:
If we do fix this, we should be careful to see how `singletons` reacts, since it turns out to be [very sensitive](https://github.com/goldfirere/singletons/issues/358#issuecomment-416089314) to the number of type variable binders that a reified class method's type signature contains at the beginning.

RyanGlScott:
A similar problem exists for data constructors:

```haskell
{-# LANGUAGE PolyKinds #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE TemplateHaskell #-}
module Main where

import Data.Proxy
import Language.Haskell.TH hiding (Type)
import Language.Haskell.TH.Desugar

main :: IO ()
main = print
 $([d| data D (a :: k) = MkD { unD :: Proxy a }
     |] >>= \ds -> do
[...]

PR #105, opened 2018-12-31
GHC 8.8 support

### LINKED ISSUE #90: GHC 8.8 checklist
This issue serves as a reminder to do certain things before we ship a new major release that supports GHC 8.8:

* [ ] Update `th-desugar` to handle `ImplicitParams` and `RecursiveDo` (see [this GHC commit](http://git.haskell.org/ghc.git/commit/9c6b7493db24977595b17046e15baf76638b5317)).
* [ ] Update `th-desugar` to handle the [More explicit foralls](http://git.haskell.org/ghc.git/commit/512eeb9bb9a81e915bfab25ca16bc87c62252064) proposal.
* [ ] Update `th-desugar` to accommodate visible kind applications.

RyanGlScott:
This adds support for:

* Visible kind application
* More explicit foralls
* Implicit params

Although `template-haskell-2.15` also adds support for `RecursiveDo`, desugaring it proves to be rather involved, so I've decided to punt on this.

Fixes #90.

### DIFF
--- a/Language/Haskell/TH/Desugar/AST.hs
+++ b/Language/Haskell/TH/Desugar/AST.hs
@@ -42,6 +42,7 @@ data DPat = DLitP Lit
 -- types and kinds.
 data DType = DForallT [DTyVarBndr] DCxt DType
            | DAppT DType DType
+           | DAppKindT DType DKind
            | DSigT DType DKind
            | DVarT Name
            | DConT Name
@@ -95,8 +96,9 @@ data DDec = DLetDec DLetDec
           | DOpenTypeFamilyD DTypeFamilyHead
           | DClosedTypeFamilyD DTypeFamilyHead [DTySynEqn]
           | DDataFamilyD Name [DTyVarBndr] (Maybe DKind)
-          | DDataInstD NewOrData DCxt Name [DType] (Maybe DKind) [DCon] [DDerivClause]
-          | DTySynInstD Name DTySynEqn
+          | DDataInstD NewOrData DCxt (Maybe [DTyVarBndr]) DType (Maybe DKind)
+                       [DCon] [DDerivClause]
+          | DTySynInstD DTySynEqn
           | DRoleAnnotD Name [Role]
           | DStandaloneDerivD (Maybe DDerivStrategy) DCxt DType
           | DDefaultSigD Name DType
@@ -245,7 +247,7 @@ data DForeign = DImportF Callconv Safety String Name DType
 data DPragma = DInlineP Name Inline RuleMatch Phases
              | DSpecialiseP Name DType (Maybe Inline) Phases
              | DSpecialiseInstP DType
-             | DRuleP String [DRuleBndr] DExp DExp Phases
+             | DRuleP String (Maybe [DTyVarBndr]) [DRuleBndr] DExp DExp Phases
              | DAnnP AnnTarget DExp
              | DLineP Int String
              | DCompleteP [Name] (Maybe Name)
@@ -257,7 +259,7 @@ data DRuleBndr = DRuleVar Name
                deriving (Show, Typeable, Data, Generic)
 
 -- | Corresponds to TH's @TySynEqn@ type (to store type family equations).
-data DTySynEqn = DTySynEqn [DType] DType
+data DTySynEqn = DTySynEqn (Maybe [DTyVarBndr]) DType DType
                deriving (Show, Typeable, Data, Generic)
 
 #if __GLASGOW_HASKELL__ < 707
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -680,40 +680,18 @@ dsDecs = concatMapM dsDec
 
 -- | Desugar a single @Dec@, perhaps producing multiple 'DDec's
 dsDec :: DsMonad q => Dec -> q [DDec]
-dsDec d@(FunD {}) = fmap (map DLetDec . fst) $ dsLetDec d
-dsDec d@(ValD {}) = fmap (map DLetDec . fst) $ dsLetDec d
+dsDec d@(FunD {}) = dsTopLevelLetDec d
+dsDec d@(ValD {}) = dsTopLevelLetDec d
 #if __GLASGOW_HASKELL__ > 710
-dsDec (DataD cxt n tvbs mk cons derivings) = do
-  tvbs'    <- mapM dsTvb tvbs
-  all_tvbs <- nonFamilyDataTvbs tvbs' mk
-  let data_type = nonFamilyDataReturnType n all_tvbs
-  (:[]) <$> (DDataD Data <$> dsCxt cxt <*> pure n
-                         <*> pure tvbs' <*> mapM dsType mk
-                         <*> concatMapM (dsCon tvbs' data_type) cons
-                         <*> mapM dsDerivClause derivings)
-dsDec (NewtypeD cxt n tvbs mk con derivings) = do
-  tvbs'    <- mapM dsTvb tvbs
-  all_tvbs <- nonFamilyDataTvbs tvbs' mk
-  let data_type = nonFamilyDataReturnType n all_tvbs
-  (:[]) <$> (DDataD Newtype <$> dsCxt cxt <*> pure n
-                            <*> pure tvbs' <*> mapM dsType mk
-                            <*> dsCon tvbs' data_type con
-                            <*> mapM dsDerivClause derivings)
+dsDec (DataD cxt n tvbs mk cons derivings) =
+  dsDataDec Data cxt n tvbs mk cons derivings
+dsDec (NewtypeD cxt n tvbs mk con derivings) =
+  dsDataDec Newtype cxt n tvbs mk [con] derivings
 #else
-dsDec (DataD cxt n tvbs cons derivings) = do
-  tvbs' <- mapM dsTvb tvbs
-  let data_type = nonFamilyDataReturnType n tvbs'
-  (:[]) <$> (DDataD Data <$> dsCxt cxt <*> pure n
-                         <*> pure tvbs' <*> pure Nothing
-                         <*> concatMapM (dsCon tvbs' data_type) cons
-                         <*> mapM dsDerivClause derivings)
-dsDec (NewtypeD cxt n tvbs con derivings) = do
-  tvbs' <- mapM dsTvb tvbs
-  let data_type = nonFamilyDataReturnType n tvbs'
-  (:[]) <$> (DDataD Newtype <$> dsCxt cxt <*> pure n
-                            <*> pure tvbs' <*> pure Nothing
-                            <*> dsCon tvbs' data_type con
-                            <*> mapM dsDerivClause derivings)
+dsDec (DataD cxt n tvbs cons derivings) =
+  dsDataDec Data cxt n tvbs Nothing cons derivings
+dsDec (NewtypeD cxt n tvbs con derivings) =
+  dsDataDec Newtype cxt n tvbs Nothing [con] derivings
 #endif
 dsDec (TySynD n tvbs ty) =
   (:[]) <$> (DTySynD n <$> mapM dsTvb tvbs <*> dsType ty)
@@ -727,10 +705,10 @@ dsDec (InstanceD over cxt ty decs) =
 dsDec (InstanceD cxt ty decs) =
   (:[]) <$> (DInstanceD <$> pure Nothing <*> dsCxt cxt <*> dsType ty <*> dsDecs decs)
 #endif
-dsDec d@(SigD {}) = fmap (map DLetDec . fst) $ dsLetDec d
+dsDec d@(SigD {}) = dsTopLevelLetDec d
 dsDec (ForeignD f) = (:[]) <$> (DForeignD <$> dsForeign f)
-dsDec d@(InfixD {}) = fmap (map DLetDec . fst) $ dsLetDec d
-dsDec d@(PragmaD {}) = fmap (map DLetDec . fst) $ dsLetDec d
+dsDec d@(InfixD {}) = dsTopLevelLetDec d
+dsDec d@(PragmaD {}) = dsTopLevelLetDec d
 #if __GLASGOW_HASKELL__ > 710
 dsDec (OpenTypeFamilyD tfHead) =
   (:[]) <$> (DOpenTypeFamilyD <$> dsTypeFamilyHead tfHead)
@@ -743,83 +721,27 @@ dsDec (FamilyD DataFam n tvbs m_k) =
   (:[]) <$> (DDataFamilyD n <$> mapM dsTvb tvbs <*> mapM dsType m_k)
 #endif
 #if __GLASGOW_HASKELL__ >= 807
-dsDec (DataInstD cxt mtvbs lhs mk cons derivings) = do
+dsDec (DataInstD cxt mtvbs lhs mk cons derivings) =
   case unfoldType lhs of
-    (ConT n, tys) -> do
-      mtvbs'     <- mapM (mapM dsTvb) mtvbs
-      tys'       <- mapM dsTypeArg tys
-      extra_tvbs <- mkExtraKindBinders mk
-      let lhs'          = applyDType (DConT n) tys'
-          all_tys       = tys' ++ map (DTANormal . dTyVarBndrToDType) extra_tvbs
-          all_tvbs      = case mtvbs' of
-                            Nothing    -> dataFamInstTvbsCompat all_tys
-                            Just tvbs' -> tvbs' ++ extra_tvbs
-          fam_inst_type = dataFamInstReturnType n all_tys
-      (:[]) <$> (DDataInstD Data <$> dsCxt cxt <*> pure mtvbs'
-                                 <*> pure lhs' <*> mapM dsType mk
-                                 <*> concatMapM (dsCon all_tvbs fam_inst_type) cons
-                                 <*> mapM dsDerivClause derivings)
-    (_, _) -> fail $ "Unexpected data instance LHS: " ++ pprint lhs
-dsDec (NewtypeInstD cxt mtvbs lhs mk con derivings) = do
+    (ConT n, tys) -> dsDataInstDec Data cxt n mtvbs tys mk cons derivings
+    (_, _)        -> fail $ "Unexpected data instance LHS: " ++ pprint lhs
+dsDec (NewtypeInstD cxt mtvbs lhs mk con derivings) =
   case unfoldType lhs of
-    (ConT n, tys) -> do
-      mtvbs'     <- mapM (mapM dsTvb) mtvbs
-      tys'       <- mapM dsTypeArg tys
-      extra_tvbs <- mkExtraKindBinders mk
-      let lhs'          = applyDType (DConT n) tys'
-          all_tys       = tys' ++ map (DTANormal . dTyVarBndrToDType) extra_tvbs
-          all_tvbs      = case mtvbs' of
-                            Nothing    -> dataFamInstTvbsCompat all_tys
-                            Just tvbs' -> tvbs' ++ extra_tvbs
-          fam_inst_type = dataFamInstReturnType n all_tys
-      (:[]) <$> (DDataInstD Newtype <$> dsCxt cxt <*> pure mtvbs'
-                                    <*> pure lhs' <*> mapM dsType mk
-                                    <*> dsCon all_tvbs fam_inst_type con
-                                    <*> mapM dsDerivClause derivings)
-    (_, _) -> fail $ "Unexpected newtype instance LHS: " ++ pprint lhs
+    (ConT n, tys) -> dsDataInstDec Newtype cxt n mtvbs tys mk [con] derivings
+    (_, _)        -> fail $ "Unexpected newtype instance LHS: " ++ pprint lhs
 #elif __GLASGOW_HASKELL__ > 710
-dsDec (DataInstD cxt n tys mk cons derivings) = do
-  tys'    <- mapM (fmap DTANormal . dsType) tys
-  all_tys <- dataFamInstTypeArgs tys' mk
-  let tvbs = dataFamInstTvbsCompat all_tys
-      fam_inst_type = dataFamInstReturnType n all_tys
-      lhs' = applyDType (DConT n) tys'
-  (:[]) <$> (DDataInstD Data <$> dsCxt cxt <*> pure Nothing
-                             <*> pure lhs' <*> mapM dsType mk
-                             <*> concatMapM (dsCon tvbs fam_inst_type) cons
-                             <*> mapM dsDerivClause derivings)
-dsDec (NewtypeInstD cxt n tys mk con derivings) = do
-  tys'    <- mapM (fmap DTANormal . dsType) tys
-  all_tys <- dataFamInstTypeArgs tys' mk
-  let tvbs = dataFamInstTvbsCompat all_tys
-      fam_inst_type = dataFamInstReturnType n all_tys
-      lhs' = applyDType (DConT n) tys'
-  (:[]) <$> (DDataInstD Newtype <$> dsCxt cxt <*> pure Nothing
-                                <*> pure lhs' <*> mapM dsType mk
-                                <*> dsCon tvbs fam_inst_type con
-                                <*> mapM dsDerivClause derivings)
+dsDec (DataInstD cxt n tys mk cons derivings) =
+  dsDataInstDec Data cxt n Nothing (map TANormal tys) mk cons derivings
+dsDec (NewtypeInstD cxt n tys mk con derivings) =
+  dsDataInstDec Newtype cxt n Nothing (map TANormal tys) mk [con] derivings
 #else
-dsDec (DataInstD cxt n tys cons derivings) = do
-  tys' <- mapM (fmap DTANormal . dsType) tys
-  let tvbs = dataFamInstTvbsCompat tys'
-      fam_inst_type = dataFamInstReturnType n tys'
-      lhs' = applyDType (DConT n) tys'
-  (:[]) <$> (DDataInstD Data <$> dsCxt cxt <*> pure Nothing
-                             <*> pure lhs' <*> pure Nothing
-                             <*> concatMapM (dsCon tvbs fam_inst_type) cons
-                             <*> mapM dsDerivClause derivings)
-dsDec (NewtypeInstD cxt n tys con derivings) = do
-  tys' <- mapM (fmap DTANormal . dsType) tys
-  let tvbs = dataFamInstTvbsCompat tys'
-      fam_inst_type = dataFamInstReturnType n tys'
-      lhs' = applyDType (DConT n) tys'
-  (:[]) <$> (DDataInstD Newtype <$> dsCxt cxt <*> pure Nothing
-                                <*> pure lhs' <*> pure Nothing
-                                <*> dsCon tvbs fam_inst_type con
-                                <*> mapM dsDerivClause derivings)
+dsDec (DataInstD cxt n tys cons derivings) =
+  dsDataInstDec Data cxt n Nothing (map TANormal tys) Nothing cons derivings
+dsDec (NewtypeInstD cxt n tys con derivings) =
+  dsDataInstDec Newtype cxt n Nothing (map TANormal tys) Nothing [con] derivings
 #endif
 #if __GLASGOW_HASKELL__ >= 807
-dsDec (TySynInstD eqn) = (:[]) <$> (DTySynInstD <$> dsTySynEqn (error "Unused") eqn)
+dsDec (TySynInstD eqn) = (:[]) <$> (DTySynInstD <$> dsTySynEqn unusedArgument eqn)
 #elif __GLASGOW_HASKELL__ >= 707
 dsDec (TySynInstD n eqn) = (:[]) <$> (DTySynInstD <$> dsTySynEqn n eqn)
 #else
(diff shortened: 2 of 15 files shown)


RyanGlScott:
Don't merge this yet—this patch assumes some changes to GHC HEAD which have yet to land. I'm only opening this PR now to get feedback on the design.

RyanGlScott:
I didn't realize this when writing this patch, but implicit params are actually desugared to some degree within GHC. (To be fair, I've never directly used implicit params in any code I've written, so I'm rather unfamiliar with them.) In `GHC.Classes`, there exists a [magic `IP` class](http://git.haskell.org/ghc.git/blob/ae4f1033cfe131fca9416e2993bda081e1f8c152:/libraries/ghc-prim/GHC/Classes.hs#l72) with the following Haddocks:

```haskell
-- | The syntax @?x :: a@ is desugared into @IP "x" a@
-- IP is declared very early, so that libraries can take
-- advantage of the implicit-call-stack feature
class IP (x :: Symbol) a | x -> a where
  ip :: a
```

Do you think it's worth emulating that desugaring here? Those Haddocks explain how `(?x :: a)` (i.e., `DImplicitParamT`) is desguared, although I'm still not entirely sure how `?x` (i.e., `DImplicitVarE`) or `let ?x = 42` (i.e., `DImplicitParamBindD`) are desugared, so if you know how those work, any insight you could provide would be appreciated.

RyanGlScott:
I've pushed an update which removes `DImplicitParamVarE` and `DImplicitParamT` in favor of the `IP`-based desugaring proposed in https://github.com/goldfirere/th-desugar/pull/105#issuecomment-450751201. I couldn't find any way to do away with `DImplicitParamBindD`, so I've left it in.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -662,120 +695,199 @@ dsDec (NewtypeD cxt n tvbs mk con derivings) = do
...
                             <*> mapM dsDerivClause derivings)
 #endif
-dsDec (TySynD n tvbs ty) =
+dsDec' _ (TySynD n tvbs ty) =
   (:[]) <$> (DTySynD n <$> mapM dsTvb tvbs <*> dsType ty)
-dsDec (ClassD cxt n tvbs fds decs) =
-  (:[]) <$> (DClassD <$> dsCxt cxt <*> pure n <*> mapM dsTvb tvbs
+dsDec' _ (ClassD cxt n tvbs fds decs) = do
+  tvbs' <- mapM dsTvb tvbs
+  (:[]) <$> (DClassD <$> dsCxt cxt <*> pure n <*> pure tvbs'
```
goldfirere: May as well just call `mapM` on this line.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -972,10 +1101,33 @@ dsRuleBndr :: DsMonad q => RuleBndr -> q DRuleBndr
...
+-- This requires a 'Name' as an argument since 'TySynEqn's did not have
+-- this information prior to GHC 8.8.
+dsTySynEqn :: DsMonad q => Name -> TySynEqn -> q DTySynEqn
+dsTySynEqn = dsTySynEqn' NotAssociated
+
+dsTySynEqn' :: DsMonad q => AssocInstInfo -> Name -> TySynEqn -> q DTySynEqn
+dsTySynEqn' aii _ (TySynEqn mtvbs lhs rhs) = do
+  lhs'  <- dsType lhs
+  rhs'  <- dsType rhs
+  tvbs' <- maybe (pure $ toposortTyVarsOfAssoc aii [lhs', rhs']) (mapM dsTvb) mtvbs
```
goldfirere: I know I advocated for always quantifying over variables during desugaring, but now I'm not so sure. The problem is that this toposorting is never guaranteed to actually work. So it's possible that a perfectly sensible user-written declaration would get transformed into one that's ill-scoped. That seems unacceptable, no? Actually, any use of toposorting (which, I know, predates this patch) seems very suspect.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -953,10 +1071,21 @@ dsPragma (SpecialiseP n ty m_inl phases) = DSpecialiseP n <$> dsType ty
...
-dsPragma (RuleP str rbs lhs rhs phases)  = DRuleP str <$> mapM dsRuleBndr rbs
+#if __GLASGOW_HASKELL__ >= 807
+dsPragma (RuleP str mtvbs rbs lhs rhs phases)
+                                         = DRuleP str <$> tvbs
+                                                      <*> mapM dsRuleBndr rbs
                                                       <*> dsExp lhs
                                                       <*> dsExp rhs
                                                       <*> pure phases
+  where
+    tvbs = maybe (pure []) (mapM dsTvb) mtvbs
```
goldfirere: This is inconsistent: we always quantify over, e.g., family instances, but we don't in RULES. Not only is it inconsistent, but it's wrong: `{-# RULES "blah" forall (x :: a). id (x :: a) = x #-}` is OK, but I imagine (can't test right now) `{-# RULES "blah" forall. forall (x :: a). id (x :: a) = x #-}` is not, because of the forall-or-nothing rule.

If we could toposort, then we could just extract the tyvars and quantify over them. But we can't toposort.

Bottom line: I think we need to just use `Maybe [DTyVarBndr]` like GHC does everywhere. That makes me sad (it's less simple) but, I believe, inevitable.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1400,3 +1560,24 @@ unravel (DAppT (DAppT DArrowT t1) t2) =
...
+-- (Either, [Int, Char])
+-- @
+unfold :: DType -> (DType, [DType])
+unfold = go []
+  where
+    go :: [DType] -> DType -> (DType, [DType])
+    go acc (DForallT _ _ ty) = go acc ty
+    go acc (DAppT ty1 ty2)   = go (ty2:acc) ty1
+    go acc (DSigT ty _)      = go acc ty
+    go acc (DAppKindT ty _)  = go acc ty
```
goldfirere: This loses information. But actually this case is impossible, because `unfold` is only called in GHC 7.6 and below. Still, it gives me the willies.

Language/Haskell/TH/Desugar/Expand.hs
```diff
@@ -78,14 +81,14 @@ expand_type ign = go []
     go args ty@(DLitT _)  = finish ty args
     go args ty@DWildCardT = finish ty args
 
-    finish :: DType -> [DType] -> q DType
-    finish ty args = return $ applyDType ty args
+    finish :: DType -> [DTypeArg] -> q DType
+    finish ty args = return $ foldl dAppTTypeArg ty args
```
goldfirere: `foldl dAppTTypeArg` should just be the new definition of `applyDType`. Any use of `applyDType` that doesn't work with `TypeArg`s is probably wrong.

Looked at remaining use sites. `nonFamilyDataReturnType` is OK. But `dataFamInstReturnType` is indeed wrong. See comments on `unfoldType`.

Language/Haskell/TH/Desugar/Expand.hs
```diff
@@ -105,8 +111,30 @@ expand_pred ign = go []
...
+
+-- | An argument to a type, either a normal type ('DTANormal') or a visible
+-- kind application ('DTyArg').
+data DTypeArg
+  = DTANormal DType
+  | DTyArg DKind
+  deriving (Data, Typeable)
+
+-- | Convert a 'DTypeArg' to a 'DType'.
+dTypeArgType :: DTypeArg -> DType
```
goldfirere: This looks very suspicious. Any usage means that we ignore the difference between `@blah` and `blah`. I've looked at use sites, and I'm worried that they're all wrong, I'm afraid.

Language/Haskell/TH/Desugar/Sweeten.hs
```diff
@@ -340,6 +401,11 @@ typeToTH DWildCardT = WildCardT
 #else
 typeToTH DWildCardT = error "Wildcards supported only in GHC 8.0+"
 #endif
+#if __GLASGOW_HASKELL__ >= 807
+typeToTH (DAppKindT t k)        = AppKindT (typeToTH t) (typeToTH k)
+#else
+typeToTH (DAppKindT {})         = error "Visible kind applications supported only in GHC 8.8+"
```
goldfirere: Mirroring the term-level treatment, we could just recur on the "function". This would work when the type argument is redundant.

Language/Haskell/TH/Desugar/Sweeten.hs
```diff
@@ -382,18 +448,20 @@ predToTH = go []
...
       | otherwise
       = ClassP n acc
+    go _ (DVarPr _)
+      = error "Template Haskell in GHC <= 7.8 does not support variable constraints."
     go _ DWildCardPr
       = error "Wildcards supported only in GHC 8.0+"
     go _ (DForallPr {})
       = error "Quantified constraints supported only in GHC 8.6+"
+    go _ (DAppKindPr {})
+      = error "Visible kind applications supported only in GHC 8.8+"
```
goldfirere: This, too, has a way forward on older GHCs.

Language/Haskell/TH/Desugar/Util.hs
```diff
@@ -235,6 +239,32 @@ unSigPred (ClassP n tys) = ClassP n (map unSigType tys)
...
+-- @
+-- Either Int Char
+-- @
+--
+-- would be unfolded to this:
+--
+-- @
+-- (Either, [Int, Char])
+-- @
+unfoldType :: Type -> (Type, [Type])
```
goldfirere: I think this function is just wrong, in that it should return a `(Type, [TypeArg])`. We cannot just drop visible kind arguments here, as that conflates the two equations of

```
type family KITF (a :: k) :: k where
  KITF @(Type -> Type) _ = Maybe
  KITF @Type _ = Int
```

I know this will cause a good deal more churn to change this type, but I think it has to be done.

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: Oops! You're right, I totally forgot to quantify here.

(As I say elsewhere, I don't understand what the issue with toposorting is that you're referring to.)

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: > The problem is that this toposorting is never guaranteed to actually work. So it's possible that a perfectly sensible user-written declaration would get transformed into one that's ill-scoped.

I don't know what you mean here. Can you give an example that would fail here?

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: We certainly could have a variant of this function that returns `(DType, [DTypeArg])` for completeness' sake. We'll still need this variant, however, since there are situations where we need to filter out the implicit arguments (e.g., the code in `L.H.T.D.Expand`).

Language/Haskell/TH/Desugar/Expand.hs
RyanGlScott: > `foldl dAppTTypeArg` should just be the new definition of `applyDType`. Any use of `applyDType` that doesn't work with `TypeArg`s is probably wrong.

Sure, although I'd certainly like to keep _something_ around with the type `DType -> [DType] -> DType`, since most of the time one doesn't care about visible kind applications.

Language/Haskell/TH/Desugar/Expand.hs
RyanGlScott: Yikes—I have no idea what I was smoking when I wrote this. Judging by this function's call sites, what I _should_ have written is a function which takes `[DTypeArg]` as input and filters out the `DTyArg`s, returning only the `DType`s from each `DTANormal`.

Language/Haskell/TH/Desugar/Sweeten.hs
RyanGlScott: Ah, I had no idea that we did this.

Language/Haskell/TH/Desugar/Sweeten.hs
RyanGlScott: Will do.

Language/Haskell/TH/Desugar/Util.hs
RyanGlScott: Makes sense.

Language/Haskell/TH/Desugar/AST.hs
```diff
@@ -84,6 +86,7 @@ data DLetDec = DFunD Name [DClause]
              | DSigD Name DType
              | DInfixD Fixity Name
              | DPragmaD DPragma
+             | DImplicitParamBindD String DExp
```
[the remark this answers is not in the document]
goldfirere: I'm thinking of something like

```
let new_x_val = ip @"y"; new_y_val = ip @"x" in
bindIP @"x" new_x_val $
bindIP @"y" new_y_val $
...
```

Note that we make both calls to `ip` *before* calling `bindIP`. A more naive treatment wouldn't do this. The names `new_x_val` and `new_y_val` are whipped up fresh.

Language/Haskell/TH/Desugar/Core.hs
goldfirere:
```
type family F (a :: (Proxy (b :: k), k))
type instance F (a :: (Proxy b, k))
```

We would, I'm sure, produce

```
type instance forall b k a. F (a :: (Proxy b, k))
```

We put the `a` last because we cleverly determine that `a` depends on `b` and `k`. But we have no way to know that `b` depends on `k`! So we get that bit wrong. :(

Language/Haskell/TH/Desugar/Core.hs
goldfirere: I suppose Core doesn't have an export list, and I was lulled into thinking `unfold` wasn't exported. If we need to filter out `@` arguments, we should have a function that does that separately. But I think this function here should return a list of `TypeArg`s.

Language/Haskell/TH/Desugar/Expand.hs
goldfirere: I would argue not to. We apply a type to arguments. And now arguments can take two possible forms. I think this will work nicely in practice... my guess is you'll have just one `foldl DAppT`, in `nonFamilyDataReturnType`. Actually, you won't even have that, because `dTyVarBndrToDType` makes more sense as `dTyVarBndrToDTypeArg`.

Language/Haskell/TH/Desugar/Expand.hs
goldfirere: That last bit -- converting a `[DTyArg]` to a `[DType]` via a filter -- makes sense in some scenarios. And, yes, I agree that it will work where you currently call `dTypeArgType`.

Language/Haskell/TH/Desugar/AST.hs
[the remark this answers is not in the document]
RyanGlScott: Makes sense. I suppose that if you had something simple like this:

```haskell
let ?x = 42 in ?x
```

Then one could desugar it in two ways:

1. The very straightforward `bindIP @"x" 42 (ip @"x")`.
2. The slightly more indirect `let new_x_val = 42 in bindIP @"x" new_x_val (ip @"x")`, as suggested by your algorithm.

Unless I adopt some sort of optimization pass, I think we'd end up producing the slightly more verbose code in option 2. Does that sound acceptable?

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: Forgive me for being slow here, but I don't see what the issue is with the type instance you've written. GHC happily accepts that instance, as far as I can tell:

```
λ> type family F (a :: (Proxy (b :: k), k))
λ> type instance forall b k a. F (a :: (Proxy b, k)) = Int
λ> :i F
type family F (a :: (Proxy b, k)) :: *
        -- Defined at <interactive>:10:1
type instance forall k (b :: k) (a :: (Proxy b, k)). F a = Int
        -- Defined at <interactive>:12:29
```

To be clear: I don't have a strong opinion on whether we use `Maybe [DTyVarBndr]` or `[DTyVarBndr]` here. I just want to set the record straight on `toposortTyVarOf`, since if it really is bogus, then large swaths of `th-desugar` are bogus, since we use `toposortTyVarsOf` in several places.

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: Fair enough. I suppose not defining a top-level function with type `DType -> (DType, [DType])` will force you to think carefully before you decide to filter out the visible kind arguments.

Language/Haskell/TH/Desugar/Expand.hs
RyanGlScott: Very well, you've convinced me that `DTypeArg` is the way of the future.

Language/Haskell/TH/Desugar/AST.hs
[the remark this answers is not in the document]
goldfirere: Yes, just do the dumb, verbose thing always. No one will ever exercise this code, anyway. :)

Language/Haskell/TH/Desugar/Core.hs
goldfirere: Urgh. We probably shouldn't accept it with the variables mis-ordered. But I suppose there's no harm in doing so.

We do use toposort in other scenarios, no? I guess not. I see it used only for type/data instance equations. I suppose if GHC is forgiving in these places, then toposorting is OK. But perhaps we can do better: *don't* toposort! That is, if GHC doesn't care about ordering, then why should we even try? Along similar lines, I see that toposorting is done in a call to `InClsInst`... but those variables are only ever used in order to remove variables from other lists, so sorting seems very superfluous.

I guess the upshot is that toposorting is never fully correct, so it seems somewhat useless.

RyanGlScott: > Urgh. We probably shouldn't accept it with the variables mis-ordered. But I suppose there's no harm in doing so.

My position is that it is not `th-desugar`'s job to emulate GHC's typechecker. `th-desugar` should certainly try its hardest to take advantage of any syntactic kind information that is provided (which is `toposortTyVarsOf`'s role), but it's never going to be as foolproof as typechecked code is going to be. And `forall b k a. <...>` is a smidgen better than `forall a b k. <...>` (without needing too much implementation effort to achieve), which I think is worthwhile.

(For what it's worth, after some further investigation, I now agree with you that GHC accepting `type instance forall b k a. F (a :: (Proxy b, k)) = Int` is rather bizarre, since something like `f :: forall b k (a :: k). SameKind a b` _is_ rejected! But that ultimately doesn't change my position—this is GHC's problem, not `th-desugar`'s.)

> That is, if GHC doesn't care about ordering, then why should we even try?

There's some benefit to sort things according to dependency order. It means that users can add sprinkle in additional kind information to improve `th-desugar`'s "inference" when desugaring. For instance, if a user finds `forall b k a. <...>` inadequate, then she can add a kind signature of `k` to the `b` in the type instance to force `th-desugar` to realize that the order should be `forall k b a. <...>`.

> Along similar lines, I see that toposorting is done in a call to `InClsInst`... but those variables are only ever used in order to remove variables from other lists, so sorting seems very superfluous.

That's a good point. We could get away with something simpler like `fvDType` there, since the order happens to be irrelevant.

> I guess the upshot is that toposorting is never fully correct, so it seems somewhat useless.

I won't claim that `toposortTyVarsOf` is perfect—after all, it can only take advantage of kind information that's expressed in a purely syntactic form. But it's certainly not useless: it absolutely fixes many bugs in `th-desugar`, such as #88 and #103. Moreover, I claim it doesn't make things any worse—the only difference between `toposortTyVarOf` and, say, `fvDType` is the order in which the free variables are returned. In light of this, I'd argue that we shouldn't let perfect be the enemy of good, and that we should use `toposortTyVarsOf` where possible to attempt to respect dependency order when it is important.

RyanGlScott: > My problem here is that `toposortTyVarsOf` has no real specification.

Sure it does! The specification is that `toposortTyVarsOf` computes the set free variables of a list of types such that:

1. Whenever an explicit kind signature of the form `(A :: K)` is encountered, the free variables of `K` will always appear to the left of the free variables of `A` in the returned result.
2. The constraint in (1) notwithstanding, free variables will appear in left-to-right order of their original appearance.

> I believe th-desugar should have this property: Given an TH entity _e_, splicing _e_ has the same effect as desugaring, sweetening, and splicing _e_, modulo text of error messages (and things like `:info`).

Unfortunately, I don't think this is a property that will ever hold 100% of the time. All of these examples have shown that given enough cleverness (and tricks like `SameKind` and `Proxy`), one will always be able to conjure up a program which `th-desugar` chokes on. And I'm fine with that—we'd need to implement type inference in Template Haskell to actually achieve this property, methinks.

> The problem with use of `toposort` is that it might take a type-correct _e_ and turn it into something that is not type-correct.

A bigger problem is that when this proposal [is implemented](https://github.com/ghc-proposals/ghc-proposals/blob/43ca1c1e65898bce973dca5ee2dcad87753ae893/proposals/0024-no-kind-vars.rst), `th-desugar`'s _existing_ treatment for things like the program in #88 will lead to type-incorrect programs, since the kind variables won't be explicitly quantified. We need to do _something_ to address this, and `toposortTyVarsOf` is the best way forward that I've seen.

> I looked at #88, and there is no use of `toposort` there.

My apologies, I was referring to `th-abstraction`'s equivalent of `toposortTyVarsOf` (that works over `Type`s instead of `DType`s), which is called `freeVariablesWellScoped` instead.

> So, I claim that we either have to give up on the principle articulated above or maintain both H98 and GADT syntaxes for constructors. If we do give up on the principle, we have two ways forward: we could continue to use `toposort`, which will sometimes produce erroneous code or reorder variables, but would be right most of the time; or we could do my refactor of `DCon`, which would never produce erroneous code but would, in rare scenarios, reorder variables w.r.t. VTA.
> 
> Of the options before us, I probably favor keeping H98 syntax, as it's the only option that's Correct. But it's annoying, so I could be convinced to live without the principle. Living without the principle, I still favor dropping `toposort`, and instead explaining that we get VTA wrong for H98 constructors.

Sigh. I really, _really_ don't want to go back to having different syntaxes for H98 and GADT constructors. An explicit design principle that we set out in #68 was to have a single syntax form for data type heads and constructors to avoid needing copies of each to represent the particular syntactic forms that are exclusive to each one (e.g., explicit return kinds in GADT heads). I really don't want to roll back this progress for the sake of making a handful of programs likelier to splice.

-----

Bottom line: `th-desugar` has a number of places where it's nice to transmogrify existing TH syntax into a different form (e.g., reified class methods, reified record selectors, data constructors, etc.) that requires determining a fully quantified type from type variable binders that possibly leave some variables implicit. In these situations, `toposortTyVarsOf`/`freeVariablesWellScoped` is a simple fix that works well in most cases (and in cases where it doesn't work, it can be made to work by simply adding more explicit kinds to guide it along). I'd greatly favor this solution over complicating the AST.

goldfirere: OK. Fine. You win. But let's put a note in the README that we don't guarantee that desugared code will type-check if there is hard-to-spot dependency among variables.

RyanGlScott: Will do! I can also put the specification of `toposortTyVarsOf` into its Haddocks to make it clearer what it is doing.

RyanGlScott:
Back to another thorny issue: `Maybe [DTyVarBndr]` vs. `[DTyVarBndr]`. After trying to use this patch with `singletons`, I'm coming to the conclusion that we should definitely use `Maybe [DTyVarBndr]`, for two reasons:

1. Using `[DTyVarBndr]` requires anyone who is constructing a type family equation to pre-compute all of the bound type variables in advance. This gets especially annoying when dealing with associated types, as I've alluded to earlier. (This means that we could get rid of this `AssocInstInfo` crap, which would be nice.)
2. Perhaps even more critically, requiring `[DTyVarBndr]` forces me to generate the following code in one place:

   ```haskell
   instance SingKind (Maybe a) where
     type forall. Demote (Maybe a) = Maybe (Demote a)
     ...
   ```

   But, due to [Trac #16116](https://ghc.haskell.org/trac/ghc/ticket/16116), that doesn't actually compile! So this actually _blocks_ me from porting over `singletons`, which is pretty dire.
3. On a more philosophical level, requiring `[DTyVarBndr]` is a bit incongruous with other AST forms in `th-desugar`. (It would be as if we required a `DForallT` at the front of every `DType`.)

Do you agree?

(Side note: even if we adopt `Maybe [DTyVarBndr]`, we'll still need to invoke `freeVariablesWellScoped` when reifying associated type class instances, for the reasons described in `Note [Rejigging reified type family equations variable binders]`.)

goldfirere:
Yes, I think all signs are pointing to using `Maybe [DTyVarBndr]`. The reason I advocated against this previously is that th-desugar is all about making things easier for downstream consumers, and a `Maybe` is harder to deal with than a list. But, in the end, I don't think we can really do it.

About the `singletons` blockage: I've wondered for some time whether `singletons` shouldn't produce TH directly, instead of th-desugar AST. After all, th-desugar was designed to be consumed easily, not produced easily. When I initially rewrote singletons to use th-desugar, I never considered outputting TH directly, but I suspect it would be a simplification. In any case, it's nothing we need to worry about today.

RyanGlScott:
I've pushed an overhaul which (I believe) addresses all of your concerns. How is this looking now?

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -705,57 +742,101 @@ dsDec (FamilyD TypeFam n tvbs m_k) = do
...
   (:[]) <$> (DDataFamilyD n <$> mapM dsTvb tvbs <*> mapM dsType m_k)
 #endif
-#if __GLASGOW_HASKELL__ > 710
+#if __GLASGOW_HASKELL__ >= 807
+dsDec (DataInstD cxt mtvbs lhs mk cons derivings) = do
+  case unfoldType lhs of
+    (ConT n, tys) -> do
+      mtvbs'     <- mapM (mapM dsTvb) mtvbs
+      tys'       <- mapM dsTypeArg tys
+      extra_tvbs <- mkExtraKindBinders mk
```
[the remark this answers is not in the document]
RyanGlScott: You are quite right. As part of my general refactoring of this code that I just pushed, I've removed this particular use of `mkExtraKindBinders`.

Note that `mkExtraDKindBinders` is used in `singletons`, so it can't be removed here. `dTyVarBndrToDType` is used elsewhere in `th-desugar` as well.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -705,57 +742,101 @@ dsDec (FamilyD TypeFam n tvbs m_k) = do
...
+          all_tvbs      = case mtvbs' of
+                            Nothing    -> dataFamInstTvbsCompat all_tys
+                            Just tvbs' -> tvbs' ++ extra_tvbs
+          fam_inst_type = dataFamInstReturnType n all_tys
+      (:[]) <$> (DDataInstD Data <$> dsCxt cxt <*> pure mtvbs'
+                                 <*> pure lhs' <*> mapM dsType mk
+                                 <*> concatMapM (dsCon all_tvbs fam_inst_type) cons
+                                 <*> mapM dsDerivClause derivings)
+    (_, _) -> fail $ "Unexpected data instance LHS: " ++ pprint lhs
+dsDec (NewtypeInstD cxt mtvbs lhs mk con derivings) = do
```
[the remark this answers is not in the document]
RyanGlScott: I've factored out `dsDataDec` and `dsDataInstDec`.

Language/Haskell/TH/Desugar/Sweeten.hs
```diff
@@ -125,38 +128,70 @@ decToTH (DDataFamilyD n tvbs mk) =
...
-decToTH (DDataInstD Newtype cxt n tys _mk [con] derivings) =
+  where
+    lhs' = typeToTH lhs
+    (n, lhs_args) =
+      case unfoldType lhs' of
+        (ConT n', lhs_args') -> (n', filterTANormals lhs_args')
+        (_, _) -> error $ "Illegal data instance LHS: " ++ pprint lhs'
+#endif
+#if __GLASGOW_HASKELL__ >= 807
+decToTH (DDataInstD Newtype cxt mtvbs lhs mk [con] derivings) =
```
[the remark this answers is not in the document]
RyanGlScott: I've factored this out into `dataInstDecToTH`.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -814,28 +901,78 @@ dsTypeFamilyHead n tvbs m_kind = do
...
+      ip_binder :: DExp -> DExp
+      ip_binder = foldr (.) id ip_binders
+  return (let_decs, ip_binder)
 
--- | Desugar a single @Dec@, perhaps producing multiple 'DLetDec's
-dsLetDec :: DsMonad q => Dec -> q [DLetDec]
+-- | Desugar a single 'Dec' that can appear in a @let@ expression.
+-- This produces the following output:
+--
+-- * One or more 'DLetDec's (a single 'Dec' can produce multiple 'DLetDec's
```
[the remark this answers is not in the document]
goldfirere: But what if the `let` binds only implicit parameters? Then the first return value would be empty, I think.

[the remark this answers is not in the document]
RyanGlScott: Again, no? As the Haddocks for `dsLetDec` explain, each implicit param binding creates two things:

* A single `DLetDec` binding something like `new_x_val` (e.g., `let new_x_val = 42`)
* A `DExp -> DExp` function to actually bind `?x` (e.g., `\z -> bindIP @"x" new_x_val z`)

So I still maintain that there will always be at least one `DLetDec` returned.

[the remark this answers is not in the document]
goldfirere: Oops. You're right. Sorry.

RyanGlScott:
Ping?

goldfirere:
If you've resolved all my objections, then go for it

PR #107, opened 2019-01-03
Fix #106 by removing DPred

### LINKED ISSUE #106: Consider abandoning `DPred`
Once upon a type, predicates were much simpler than ordinary types. In those halcyon days, it made sense to separate `DPred` from `DType` so that code that consumes predicates needn't match on impossible cases.

Those days are long gone now. It's true that we'll never have, say, a literal at the head of a predicate, but that's really a kind error, not a syntax error. And I worry that keeping `DPred` separate isn't really worth it anymore.

What think you?

goldfirere:
This patch also changes the constructor names for `DPat`, which
no longer need `Pa` suffixes, but can suffice with `P`.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -12,6 +12,9 @@ Version 1.10
 * Locally reified class methods, data constructors, and record selectors now
   quantify kind variables properly.
 * Add more functions which compute free variables.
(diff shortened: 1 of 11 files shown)


goldfirere:
This has been merged.

PR #115, opened 2019-02-27
Make th-desugar deterministic

### LINKED ISSUE #112: toposortTyVarsOf depends on the order of gensymmed unique values for Names
What does this program do?

```haskell
module Bug where

import Language.Haskell.TH

main :: IO ()
main =
  $(do a <- newName "a"
       b <- newName "b"
       runIO $ do putStrLn "toposortTyVarsOf [a,b]"
                  print $ toposortTyVarsOf [DVarT a, DVarT b]
       runIO $ do putStrLn "toposortTyVarsOf [b,a]"
                  print $ toposortTyVarsOf [DVarT b, DVarT a]
       [| return () |])
```

It turns out that both invocations of `toposortTyVarsOf` will return the _same_ answer:

```
$ ghc -package-db ~/.cabal/store/ghc-8.6.3/package.db/ Bug.hs -fforce-recomp
toposortTyVarsOf [a,b]
[DPlainTV a_6989586621679035542,DPlainTV b_6989586621679035543]
toposortTyVarsOf [b,a]
[DPlainTV a_6989586621679035542,DPlainTV b_6989586621679035543]
```

Yikes! This explicitly goes against the documentation for `toposortTyVarsOf`, which guarantees that it preserves the left-to-right ordering of variables that do not depend on each other.

So why is this happening? When we create a name with `newName`, it is bundled with a gensymmed unique number. The `Ord Name` instance prioritizes this unique over everything else, so if you have `Name`s `a_1` and `b_0` (with respective uniques of 1 and 0), then sorting them according to their `Ord` instance would yield `{b_0, a_1}` (despite the fact that `a` appears before `b` lexicographically).

It turns out that deep within the bowels of `toposortTyVarsOf`, we're constructing a graph using the [`graphFromEdges`](http://hackage.haskell.org/package/containers-0.6.0.1/docs/Data-Graph.html#v:graphFromEdges) function, which uses `Ord` internally! I don't know how I overlooked this originally, but this is the source of all of our troubles.

To further emphasize how dangerous `toposortTyVarsOf`'s current behavior is, if you compile the program above with `-dunique-increment=-1`, you'll get a completely different answer:

```
$ ghc -package-db ~/.cabal/store/ghc-8.6.3/package.db/ Bug.hs -fforce-recomp -dunique-increment=-1
toposortTyVarsOf [a,b]
[DPlainTV b_7061644215716911977,DPlainTV a_7061644215716911978]
toposortTyVarsOf [b,a]
[DPlainTV b_7061644215716911977,DPlainTV a_7061644215716911978]
```

~~Patch incoming.~~ (**EDIT**: See https://github.com/goldfirere/th-desugar/issues/112#issuecomment-461990221.)

RyanGlScott:
There were several places in `th-desugar` that would compute entirely different things depending on the order in which GHC's unique numbers happened to be computed:

* Previously, `toposortTyVarsOf` was performing a topological sort on a graph constructed from the `graphFromEdges` function in the `containers` library, which relies on the `Ord Name` instance to work. The `Ord Name` instance sorts first by the unique value stored in the `Name`, making this instance nondeterministic should its behavior ever leak out to user-facing code.

  In this patch, I `graphFromEdges` component of `toposortTyVarsOf` with `scopedSort`, a deterministic function which mimics the topological sorting without relying on the `Ord Name` instance. `scopedSort`'s implementation is cargo-culted from the function of
  the same name in the GHC source code.
* All functions in `th-desugar` that compute free variables were returning a `Set Name`. This becomes dangerous since other parts of the codebase (in `th-desugar` and `singletons`) call `toList` on these `Set`s of `Name`s, converting them into a list sorted in ascending order by their `Name`s!

  In this patch, I introduce `OMap` and `OSet` data structures that behave like `Map` and `Set`, but remembering the order in which elements were inserted. Their implementations are mostly taken from the `ordered-containers` library. I didn't depend on `ordered-containers` directly since I needed to make some minor modifications (among which are https://github.com/dmwit/ordered-containers/pull/7, https://github.com/dmwit/ordered-containers/pull/8, https://github.com/dmwit/ordered-containers/pull/9, and https://github.com/dmwit/ordered-containers/pull/5) to suit `th-desugar`'s needs, and the maintainer of `ordered-containers` does not appear to be responsive to pull requests.

  I replaced every nondeterministic use of `Map` and `Set` with `OMap` and `OSet`, respectively. Note that not every use of `Map` and `Set` needs to be replaced—for instance, the code in `Language.Haskell.TH.Desugar.Subst` uses `Map`s to represent substitutions, but this is fine, since the internal order of the `Map`s is never leaked to the user.

This diff appears large, but most of the modifications are routine changes brought about by switching data structures. Now that these changes are in place, we can finally run the `th-desugar` test suite with `-dunique-increment=-1` and have it still pass! Hooray!

Fixes #112.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -48,6 +48,8 @@ Version 1.10
   of type variable binders in `forall`s.
 * Fix a bug in which `getRecordSelectors` would omit record selectors from
   GADT constructors.
+* Fix a bug in which `toposortTyVarsOf` would sometimes not preserve
+  the left-to-right ordering of `Name`s generated with `qNewName`.
 * Locally reified class methods, data constructors, and record selectors now
   quantify kind variables properly.
 * Desugared ADT constructors now quantify kind variables properly.
--- a/Language/Haskell/TH/Desugar.hs
+++ b/Language/Haskell/TH/Desugar.hs
@@ -112,16 +112,18 @@ module Language.Haskell.TH.Desugar (
 
 import Language.Haskell.TH.Desugar.AST
 import Language.Haskell.TH.Desugar.Core
-import Language.Haskell.TH.Desugar.FV
-import Language.Haskell.TH.Desugar.Util
-import Language.Haskell.TH.Desugar.Sweeten
-import Language.Haskell.TH.Syntax
-import Language.Haskell.TH.Desugar.Reify
 import Language.Haskell.TH.Desugar.Expand
+import Language.Haskell.TH.Desugar.FV
 import Language.Haskell.TH.Desugar.Match
+import qualified Language.Haskell.TH.Desugar.OSet as OS
+import Language.Haskell.TH.Desugar.Reify
 import Language.Haskell.TH.Desugar.Subst
+import Language.Haskell.TH.Desugar.Sweeten
+import Language.Haskell.TH.Desugar.Util
+import Language.Haskell.TH.Syntax
 
 import Control.Monad
+import qualified Data.Foldable as F
 import Data.Function
 import Data.List
 import qualified Data.Map as M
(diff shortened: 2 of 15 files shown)


goldfirere:
I'm happy with this direction of travel. Is this in want of specific feedback somewhere?

PR #118, opened 2019-05-21
Add `Maybe [DTyVarBndr]` fields to `DInstanceD`/`DStandaloneDerivD` (#117)

### LINKED ISSUE #117: Should DInstanceD have a Maybe [DTyVarBndr] field?
This is the current definition of `DInstanceD`:

https://github.com/goldfirere/th-desugar/blob/e420eaeb80f108c186c26ea706c7d80ce5e3d106/Language/Haskell/TH/Desugar/AST.hs#L94

This makes sense, as it mirrors `InstanceD` in `template-haskell`. However, I'm wondering if the API might be made more complete by adding an additional field to `DInstanceD` of type `Maybe [DTyVarBndr]`. This idea was born out of a realization I had when I thought about instances like the one below (inspired by [GHC #16518](https://gitlab.haskell.org/ghc/ghc/issues/16518)):

```hs
class HTestEquality (f :: forall k. k -> Type) where
  hTestEquality :: forall k1 k2 (a :: k1) (b :: k2).
                   f a -> f b -> Maybe (a :~~: b)
newtype T :: (forall k. k -> Type) -> (forall k. k -> Type) where
  MkT :: forall (f :: forall k. k -> Type) k (a :: k). f a -> T f a

instance forall (f :: forall k. k -> Type). HTestEquality f => HTestEquality (T f) where
  hTestEquality (MkT x) (MkT y) = hTestEquality x y
```

How exactly would one represent `instance forall (f :: forall k. k -> Type). HTestEquality f => HTestEquality (T f)` in Template Haskell? There's not an obvious place to put the `forall (f :: forall k. k -> Type)` part, and if we omit it, then the instance will fail to typecheck. Luckily, there _is_ a way to accomplish this: use the following TH AST:

```
λ> ; $(newName "f" >>= \f -> newName "k" >>= \k -> pure [InstanceD Nothing [] (ForallT [KindedTV f (ForallT [PlainTV k] [] (ArrowT `AppT` VarT k `AppT` StarT))] [ConT ''HTestEquality `AppT` VarT f] (ConT ''HTestEquality `AppT` (ConT ''T `AppT` VarT f))) []])
<interactive>:33:5-254: Splicing declarations
    newName "f"
      >>=
        \ f_ahqw
          -> newName "k"
               >>=
                 \ k_ahqx
                   -> pure
                        [InstanceD
                           Nothing
                           []
                           (ForallT
                              [KindedTV
                                 f_ahqw
                                 (ForallT
                                    [PlainTV k_ahqx] [] (ArrowT `AppT` VarT k_ahqx `AppT` StarT))]
                              [ConT ''HTestEquality `AppT` VarT f_ahqw]
                              (ConT ''HTestEquality `AppT` (ConT ''T `AppT` VarT f_ahqw)))
                           []]
  ======>
    instance forall (f_ahrt :: forall k_ahru.
                               k_ahru -> Data.Kind.Type).
             HTestEquality f_ahrt => HTestEquality (T f_ahrt)
```

This trick is a bit grimy, but it has the nice property that we don't have to change the TH AST to accommodate the (relatively obscure) use case of higher-ranked kind variables. Instead, we only need make this change in the downstream `th-desugar` library by adding a field of type `Maybe [DTyVarBndr]` to `DInstanceD`. If that field is instantiated with `Just tvs`, then we use the trick above and explicitly quantify `tvs`. If it's `Nothing`, then proceed like we normally do.

If we extend `DInstanceD`, then we should also give `DStandaloneDerivD` the same treatment.

RyanGlScott:
This resolves #117 by implementing the trick described in that issue. Alas, this trick does not work on all versions of GHC that `th-desugar` currently supports, so some CPP hackery is employed to produce slightly more sensible error messages in the event that the trick is used on incompatible GHC versions.

### DIFF
--- a/Language/Haskell/TH/Desugar/Sweeten.hs
+++ b/Language/Haskell/TH/Desugar/Sweeten.hs
@@ -116,6 +116,7 @@ decToTH (DInstanceD over mtvbs _cxt _ty decs) =
 #if __GLASGOW_HASKELL__ < 800 || __GLASGOW_HASKELL__ >= 802
                                 ([], DForallT _tvbs _cxt _ty)
 #else
+                                -- See #117
                                 error $ "Explicit foralls in instance declarations "
                                      ++ "are broken on GHC 8.0."
 #endif
@@ -173,6 +174,7 @@ decToTH (DStandaloneDerivD mds mtvbs _cxt _ty) =
 #if __GLASGOW_HASKELL__ < 710 || __GLASGOW_HASKELL__ >= 802
                                 ([], DForallT _tvbs _cxt _ty)
 #else
+                                -- See #117
                                 error $ "Explicit foralls in standalone deriving declarations "
                                      ++ "are broken on GHC 7.10 and 8.0."
 #endif
--- a/Language/Haskell/TH/Desugar/Sweeten.hs
+++ b/Language/Haskell/TH/Desugar/Sweeten.hs
@@ -107,12 +107,17 @@ decToTH (DDataD Newtype cxt n tvbs _mk [con] derivings) =
 decToTH (DTySynD n tvbs ty) = [TySynD n (map tvbToTH tvbs) (typeToTH ty)]
 decToTH (DClassD cxt n tvbs fds decs) =
   [ClassD (cxtToTH cxt) n (map tvbToTH tvbs) fds (decsToTH decs)]
-#if __GLASGOW_HASKELL__ >= 711
-decToTH (DInstanceD over cxt ty decs) =
-  [InstanceD over (cxtToTH cxt) (typeToTH ty) (decsToTH decs)]
+decToTH (DInstanceD over mtvbs _cxt _ty decs) =
+  [instanceDToTH over cxt' ty' decs]
+  where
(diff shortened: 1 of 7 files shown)


Language/Haskell/TH/Desugar/Sweeten.hs
```diff
@@ -159,18 +164,22 @@ decToTH (DClosedTypeFamilyD (DTypeFamilyHead n tvbs frs _ann) eqns) =
...
+decToTH (DStandaloneDerivD mds mtvbs _cxt _ty) =
+  [standaloneDerivDToTH mds cxt' ty']
+  where
+  (cxt', ty') = case mtvbs of
+                  Nothing    -> (_cxt, _ty)
+                  Just _tvbs ->
+#if __GLASGOW_HASKELL__ < 710 || __GLASGOW_HASKELL__ >= 802
+                                ([], DForallT _tvbs _cxt _ty)
+#else
+                                error $ "Explicit foralls in standalone deriving declarations "
```
goldfirere: Perhaps refer back to the ticket so readers can learn more about what the problem is?

PR #120, opened 2019-07-05
 Clean up L.H.T.D.{OMap,OSet} by using the ordered-containers package

RyanGlScott:
`ordered-containers-0.2.2` (recently released to Hackage) contains enough features that we no longer need to vendor in a modified version of the library in `th-desugar` itself. This allows us to drastically simplify the code in `L.H.T.D.{OMap,OSet}`.

There's one catch: `ordered-containers-0.2.2` only supports GHC 7.8 or later, so I had to drop `th-desugar` support for GHC 7.6. GHC 7.6 pretty old by now, so I don't think this is a huge loss. As an added bonus, this lets us delete lots of 7.6-specific CPP.


goldfirere:
Full steam ahead. Thanks!

PR #125, opened 2019-09-02
Support visible dependent quantification

### LINKED ISSUE #113: Split up DForallT into two different constructors
This is the current definition of `DForallT`:

https://github.com/goldfirere/th-desugar/blob/9b9a6f0bdb6becd3f5ae896ddb539069131f7ba0/Language/Haskell/TH/Desugar/AST.hs#L43

This isn't too surprising, since it mirrors the exact structure of `template-haskell`'s `ForallT`. But `ForallT`/`DForallT` are a bit strange, since they combine the distinct concepts of type variable quantification and context arguments. This isn't so bad on its own, since if you want to represent a type that _only_ quantified type variables, you can just leave off the context (e.g., `forall a. a` becomes `DForallT [DPlainTV a] [] (DVarT a)` in `th-desugar`), and similarly if you want to represent a type that only has a context. It's a bit annoying to have to provide an empty list of type variable binders/predicates, but it's not the end of the world.

GHC 8.10 will throw an additional wrinkle into this story, since it introduces visible dependent quantification. A type like `forall a. forall b -> Either a b` would be represented in Template Haskell as:

```haskell
ForallT [PlainTV a] [] $ ForallVisT [PlainTV b]
                       $ ConT ''Either `AppT` VarT a `AppT` VarT b
```

Notice that there's an entirely distinct `ForallVisT` constructor. It's mighty similar to `ForallT`, however, and one might wonder why Template Haskell didn't just combine `ForallT` and `ForallVisT` into a single constructor. Ultimately, Template Haskell's design choice here was motivated by two different factors:

1. `ForallT` bundles a context alongside its type variable binders. It feels somewhat strange to bundle a context with `ForallVisT`, since the type variable binders are visible but the context isn't. TH could have removed the context part and put it into its own constructor, but that would have required a breaking change.
2. In order to distinguish invisible and visible `forall`s, TH would need to bundle some sort of `ForallVisFlag` in the new `ForallT` constructor to tell them apart. But that would require... you guessed it, another breaking change.

In short, changing the design here would involve massive breaking changes for a large chunk of TH clients, so Template Haskell compromised by just introducing a separate `ForallVisT` constructor.

Fortunately, we can strive for a better design in `th-desugar` since backwards compatibility between `th-desugar` versions isn't a top priority. I propose that we split up the current `DForallT` constructor into two different constructors:

```haskell
data DType
  = DForallT DForallVisFlag [DTyVarBndr] DType
  | DCxtT DCxt DType 
  | ...

data DForallVisFlag
  = DForallVis
  | DForallInvis
```

Now visible and invisible `forall`s are unified into a single constructor, and context arguments are confined to a separate constructor.

RyanGlScott:
This adds support for visible dependent quantification (new in GHC 8.10). There are two main changes involved in this:

* The existing `DForallT` constructor is now split into two constructors, `DForallT` (for type variable binders) and `DConstrainedT` (for constraints).

  The process of desugaring and sweetening these constructors is somewhat tricky. See `Note [Desugaring and sweetening ForallT]` in `L.H.T.Desugar.Core` for the full story.
* The `unravel` function now chunks up function arguments using an auxiliary `DFunArgs` data type. We also need a version of this for `Type`s, so there is now both `unravelType` and `unravelDType` (the existing `unravel` function was renamed to `unravelDType`).

See the `CHANGES.md` file for the full list of changes.

Fixes #113.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,45 @@
 `th-desugar` release notes
 ==========================
 
+Version 1.11 [????.??.??]
+-------------------------
+* Support GHC 8.10.
+* Add support for visible dependent quantification. As part of this change,
+  the way `th-desugar` represents `forall` and constraint types has been
+  overhauled:
+  * The existing `DForallT` constructor has been split into two smaller
+    constructors:
+
+    ```diff
+     data DType
+       = ...
+    -  | DForallT [DTyVarBndr] DCxt DType
+    +  | DForallT ForallVisFlag [DTyVarBndr] DType
+    +  | DConstrainedT DCxt DType
+       | ...
+
+    +data ForallVisFlag
+    +  = ForallVis
+    +  | ForallInvis
+    ```
+
+    The previous design combined `forall`s and constraints into a single
+    constructor, while the new design puts them in distinct constructors
+    `DForallT` and `DConstrainedT`, respectively. The new `DForallT`
+    constructor also has a `ForallVisFlag` field to distinguish invisible
+    `forall`s (e.g., `forall a. a`) from visible `forall`s (e.g.,
+    `forall a -> a`).
+  * The `unravel` function has been renamed to `unravelDType` and now returns
+    `(DFunArgs, DType)`, where `DFunArgs` is a data type that represents
+    the possible arguments in a function type (see the Haddocks for `DFunArgs`
+    for more details). There is also an `unravelDType` counterpart for `Type`s
+    named `unravelType`, complete with its own `FunArgs` data type.
+
+    `{D}FunArgs` also have some supporting operations, including
+    `filter{D}VisFunArgs` (to obtain only the visible arguments) and
+    `ravel{D}Type` (to construct a function type using `{D}FunArgs` and
+    a return `{D}Type`).
+
 Version 1.10
 ------------
 * Support GHC 8.8. Drop support for GHC 7.6.
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -16,7 +16,6 @@ import Prelude hiding (mapM, foldl, foldr, all, elem, exp, concatMap, and)
 
 import Language.Haskell.TH hiding (match, clause, cxt)
 import Language.Haskell.TH.Syntax hiding (lift)
-import Language.Haskell.TH.Datatype ( resolveTypeSynonyms )
 
 #if __GLASGOW_HASKELL__ < 709
 import Control.Applicative
@@ -817,22 +816,6 @@ dsDataInstDec nd cxt n mtvbs tys mk cons derivings = do
                            <*> concatMapM (dsCon h98_tvbs h98_fam_inst_type) cons
                            <*> mapM dsDerivClause derivings)
 
--- Like mkExtraDKindBinders, but accepts a Maybe Kind
--- argument instead of DKind.
-mkExtraKindBinders :: DsMonad q => Maybe Kind -> q [DTyVarBndr]
-mkExtraKindBinders =
-  maybe (pure (DConT typeKindName)) (runQ . resolveTypeSynonyms >=> dsType)
-    >=> mkExtraDKindBinders'
-
--- | Like mkExtraDKindBinders, but assumes kind synonyms have been expanded.
-mkExtraDKindBinders' :: Quasi q => DKind -> q [DTyVarBndr]
-mkExtraDKindBinders' =
-  mkExtraKindBindersGeneric unravelDType filterDVisFunArgs DKindedTV elim_dvfa
-  where
-    elim_dvfa :: (DTyVarBndr -> r) -> (DKind -> r) -> DVisFunArg -> r
-    elim_dvfa dep _    (DVisFADep tvb) = dep tvb
-    elim_dvfa _   anon (DVisFAAnon k)  = anon k
-
 #if __GLASGOW_HASKELL__ > 710
 -- | Desugar a @FamilyResultSig@
 dsFamilyResultSig :: DsMonad q => FamilyResultSig -> q DFamilyResultSig
@@ -1303,7 +1286,8 @@ dsPred (ImplicitParamT n t) = do
   return [DConT ''IP `DAppT` DLitT (StrTyLit n) `DAppT` t']
 #endif
 #if __GLASGOW_HASKELL__ >= 809
-dsPred (ForallVisT tvbs p) = dsForallPred tvbs [] p
+dsPred t@(ForallVisT {}) =
+  impossible $ "Visible dependent quantifier seen as head of constraint: " ++ show t
 #endif
 
 -- | Desugar a quantified constraint.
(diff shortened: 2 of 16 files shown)


RyanGlScott:
See goldfirere/singletons#406 for the corresponding `singletons` PR.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1298,13 +1302,34 @@ dsPred (ImplicitParamT n t) = do
   t' <- dsType t
   return [DConT ''IP `DAppT` DLitT (StrTyLit n) `DAppT` t']
 #endif
+#if __GLASGOW_HASKELL__ >= 809
+dsPred (ForallVisT tvbs p) = dsForallPred tvbs [] p
```
goldfirere: Eep! Is this even possible? I hope not.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1298,13 +1302,34 @@ dsPred (ImplicitParamT n t) = do
...
+
+-- | Desugar a quantified constraint.
+dsForallPred :: DsMonad q => [TyVarBndr] -> Cxt -> Pred -> q DCxt
+dsForallPred tvbs cxt p = do
+  ps' <- dsPred p
+  case ps' of
+    [p'] -> (:[]) <$> (mkDForallConstrainedT ForallInvis
+                         <$> mapM dsTvb tvbs <*> dsCxt cxt <*> pure p')
+    _    -> fail "Cannot desugar constraint tuples in the body of a quantified constraint"
+              -- See Trac #15334.
```
goldfirere: s/Trac/GHC/ or s/Trac/GitLab/

CHANGES.md
```diff
@@ -1,6 +1,45 @@
...
+  the way `th-desugar` represents `forall` and constraint types has been
+  overhauled:
+  * The existing `DForallT` constructor has been split into two smaller
+    constructors:
+
+    ```diff
+     data DType
+       = ...
+    -  | DForallT [DTyVarBndr] DCxt DType
+    +  | DForallT ForallVisFlag [DTyVarBndr] DType
```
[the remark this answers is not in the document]
goldfirere: Oh, I suppose `f :: forall a. forall b. ...` is different than `f :: forall a b. blah` with `-XScopedForalls`. Urgh.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1620,3 +1696,56 @@ extractTvbKind (DKindedTV _ k) = Just k
...
+* `DForallT DForallInvis tvbs (DConstrainedT ctxt ty)` should sweeten to `ForallT tvbs ctxt ty`.
+* `DForallT DForallInvis []   (DConstrainedT ctxt ty)` should sweeten to `ForallT [] ctxt ty`.
+* `DForallT DForallInvis tvbs (DConstrainedT [] ty)`   should sweeten to `ForallT tvbs [] ty`.
+* `DForallT DForallInvis []   (DConstrainedT [] ty)`   should sweeten to `ForallT [] [] ty`.
+* For all other cases, just straightforwardly sweeten
+  `DForallT DForallInvis tvbs ty` to `ForallT tvbs [] ty` and
+  `DConstrainedT ctxt ty` to `ForallT [] ctxt ty`.
+
+One gotcha of this specification is the th-desugar roundtripping will turn the
+type `() => a` into `forall a. () => a`. In practice, Template Haskell won't
```
goldfirere: You mean `forall. () => a` there, don't you? Otherwise, deep magic must be happening.

Really, you're just observing that TH can't spot the difference between `() => a` and `forall. () => a`.

Actually, with that in mind, I don't understand this claim. What is the difference (in the TH AST) between your "from" and your "to"?

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1620,3 +1696,56 @@ extractTvbKind (DKindedTV _ k) = Just k
 -- version of 'undefined'.)
 unusedArgument :: a
 unusedArgument = error "Unused"
+
+{-
+Note [Desugaring and sweetening ForallT]
```
goldfirere: Great note. Thanks.

Language/Haskell/TH/Desugar/Sweeten.hs
```diff
@@ -497,10 +517,27 @@ predToTH DWildCardT  = WildCardT
...
+-- so that we may collapse them into a single ForallT when sweetening.
+-- See Note [Desugaring and sweetening ForallT] in L.H.T.Desugar.Core.
+predToTH (DForallT ForallInvis tvbs (DConstrainedT ctxt p)) =
+  ForallT (map tvbToTH tvbs) (map predToTH ctxt) (predToTH p)
+predToTH (DForallT fvf tvbs p) =
+  case fvf of
+    ForallInvis -> ForallT tvbs' [] p'
+    ForallVis ->
+#if __GLASGOW_HASKELL__ >= 809
+      ForallVisT tvbs' p'
```
goldfirere: I'm scared. I'm very, very scared.

Language/Haskell/TH/Desugar/Util.hs
```diff
@@ -207,27 +209,105 @@ splitTuple_maybe t = go [] t
...
   -> kind -> q [tyVarBndr]
-mkExtraKindBindersGeneric unravel mkKindedTV k = do
-  let (_, _, args, _) = unravel k
-  names <- replicateM (length args) (qNewName "a")
-  return (zipWith mkKindedTV names args)
-
--- | Decompose a function 'Type' into its type variables, its context, its
--- argument types, and its result type.
-unravelType :: Type -> ([TyVarBndr], [Pred], [Type], Type)
+mkExtraKindBindersGeneric unravel filter_vis_fun_args
```
goldfirere: I won't push hard here, but this Very Clever type seems overwrought, for such a small function. Would it be simpler just to inline? I suppose it might be simpler, but also more error-prone.

I'd actually be more amenable to a tiny type class to do this -- seems like it would be much simpler conceptually.  But I'm not set against accepting this as is.

RyanGlScott: On second thought, I agree that this is overwrought. I've simply inlined the implementations into both specializations of the function.

Well, I say "both", but it turns out there are a handful of other variations of `mkExtraKindBindersGeneric` lying around that aren't even being used! I went ahead and removed those as well. The end result is much cleaner, and the resulting code duplication isn't _that_ bad.

Language/Haskell/TH/Desugar/Core.hs
goldfirere: Right. I guess all it needs to be clear to me is to note that you're round-tripping from Haskell to th-desugar and back. Note that this round trip is generally not safe. (For example, `(x,y)` becomes `(,) x y`.) Though in other scenarios, the change preserves semantics. Here it might not... yet this lack of preservation would happen even if the round-trip only went through TH and didn't touch th-desugar, right? So I guess I'm still not exactly sure what you're trying to say here.

RyanGlScott: Indeed, this warning only makes sense if TH roundtripping actually preserved the semantics of top-level `forall`s, which it doesn't. In light of this, perhaps it's better just not to mention it at all. I'll just remove this paragraph, since it seems to cause more confusion than actual insight.

Issue #114, opened 2019-02-22
GHC 8.10 checklist


RyanGlScott:
This issue serves as a reminder to do certain things before we ship a new major release that supports GHC 8.10:

* [x] Support visible dependent quantification (see also #113).
* [x] Support standalone kind signatures and `reifyTypeWithLocals`

RyanGlScott:
Fixed with the Hackage release of [`th-desugar-1.11`](http://hackage.haskell.org/package/th-desugar-1.11).

PR #126, opened 2019-09-26
Support standalone kind signatures, add reifyTypeWithLocals

RyanGlScott:
This patch:

1. Adds support for desugaring and sweetening standalone kind signatures. This is quite straightforward, as this simply requires adding a new `DKiSigD` declaration and installing the corresponding plumbing.
2. Adds support for reifying type/kind signatures in local declarations. This is accomplished with the new `dsReifyType`, `reifyTypeWithLocals_maybe`, and `reifyTypeWithLocals` functions in `L.H.T.Desugar.Reify`. This is by far the largest part of the patch, and I've added a number of test cases to increase confidence that they are implemented correctly.

Addresses one bullet point of #114.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -119,10 +119,6 @@ install:
   - |
     echo "packages: ." >> cabal.project
   - |
-    echo "source-repository-package"                             >> cabal.project
-    echo "  type:     git"                                       >> cabal.project
-    echo "  location: https://github.com/RyanGlScott/th-orphans" >> cabal.project
-    echo "  tag:      60135a6149ceff425bdcff418ff2f4a07e7bc44d"  >> cabal.project
   - "for pkg in $($HCPKG list --simple-output); do echo $pkg | sed 's/-[^-]*$//' | (grep -vE -- '^(th-desugar)$' || true) | sed 's/^/constraints: /' | sed 's/$/ installed/' >> cabal.project.local; done"
   - cat cabal.project || true
   - cat cabal.project.local || true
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -119,6 +119,10 @@ install:
   - |
     echo "packages: ." >> cabal.project
   - |
+    echo "source-repository-package"                             >> cabal.project
+    echo "  type:     git"                                       >> cabal.project
+    echo "  location: https://github.com/RyanGlScott/th-orphans" >> cabal.project
+    echo "  tag:      60135a6149ceff425bdcff418ff2f4a07e7bc44d"  >> cabal.project
```
RyanGlScott: Currently, this will only build on GHC 8.9 if I make some upstream changes to `th-orphans` so that `qReifyType` is defined in the `Quasi` instance for `ReaderT`. I'm not sure what exactly to do about this. Should we:

1. Try getting these changes merged upstream in `th-orphans` and wait until a new release of `th-desugar` emerges before landing this patch?
2. Just vendor in this fork of `th-orphans` for now, and remove it at some point before the next `th-desugar` release?
3. Something else?

goldfirere: I vote (1). If you supply a PR for th-orphans, I would imagine it would get accepted fairly quickly.

Language/Haskell/TH/Desugar/Sweeten.hs
```diff
@@ -185,6 +185,12 @@ decToTH (DPatSynSigD n ty)        = [PatSynSigD n (typeToTH ty)]
 decToTH DPatSynD{}    = patSynErr
 decToTH DPatSynSigD{} = patSynErr
 #endif
+#if __GLASGOW_HASKELL__ >= 809
+decToTH (DKiSigD n ki) = [KiSigD n (typeToTH ki)]
+#else
+decToTH (DKiSigD {})   =
+  error "Standalone kind signatures supported only in GHC 8.10+"
```
goldfirere: We could, if someone really wanted to, try to manufacture a CUSK from a standalone signature. But let's not.

.travis.yml
RyanGlScott: `th-orphans-0.13.9` now includes definitions for `qReifyType`. I've pushed a commit which requires the test suite to use this version of `th-orphans` as the minimum.

PR #144, opened 2020-06-29
Make it depend on semigroups/fail only on GHC < 8.0

felixonmars:
They are not needed on newer GHC.

### DIFF
--- a/th-desugar.cabal
+++ b/th-desugar.cabal
@@ -47,15 +47,16 @@ library
       ghc-prim,
       template-haskell >= 2.9 && < 2.17,
       containers >= 0.5,
-      fail == 4.9.*,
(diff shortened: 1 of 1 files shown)


RyanGlScott:
I don't have any particular objection to this patch, but I have to wonder: is it really worth going through the effort of implementing conditional dependencies here? After all, neither `fail` nor `semigroups` define any code on GHC 8.0 or later, so in some sense, these dependencies are "free" on later GHCs.

felixonmars:
@RyanGlScott We can get rid of two os packages when packaging for Linux distros. Currently this is implemented as a downstream patch and I would like to get it uptsream.

RyanGlScott:
Very well, I'll land this.

PR #145, opened 2020-08-20
Final template-haskell-2.17–related changes (linear types and QualifiedDo)

### LINKED ISSUE #143: Adapt to linear types in th-desugar
How should `th-desugar` cope with a brave new linear world? At first glance, the changes that the [linear types patch](https://gitlab.haskell.org/ghc/ghc/-/commit/40fa237e1daab7a76b9871bb6c50b953a1addf23) makes to `template-haskell` might seem deceptively small:

```diff
@@ -2465,6 +2468,7 @@ data Type = ForallT [TyVarBndr Specificity] Cxt Type -- ^ @forall \<vars\>. \<ct
           | UnboxedTupleT Int             -- ^ @(\#,\#), (\#,,\#), etc.@
           | UnboxedSumT SumArity          -- ^ @(\#|\#), (\#||\#), etc.@
           | ArrowT                        -- ^ @->@
+          | MulArrowT                     -- ^ @FUN@
           | EqualityT                     -- ^ @~@
           | ListT                         -- ^ @[]@
           | PromotedTupleT Int            -- ^ @'(), '(,), '(,,), etc.@
```

The ramifications for `th-desugar` could be larger, however. Some questions we must answer are:

1. How should `DType` be changed? We could simply add a new `DMulArrowT` constructor alongside `DArrowT`, but that might fall afoul of the general `th-desugar` principle of making the AST "minimalist". Perhaps a better design is possible.
2. How should `DFunArgs` be changed? Currently, we have:

   ```hs
   data FunArgs
     = ...
     | DFAAnon DType DFunArgs
   ```

   `DFAAnon` doesn't record whether the argument to a function arrow is linear or not. We probably should, somehow.
3. How should `L.H.TH.Desugar.Reify` deal with reifying the types of data constructors? Haskell98 constructors and record GADT constructors are easy, since they will always have linear argument types (e.g., `Just :: a #-> Maybe a`). The tricky case will be prefix GADT constructors, which can have mixed linear and unrestricted arguments. To make things worse, GHC makes it difficult to know which are which. See [GHC#18378](https://gitlab.haskell.org/ghc/ghc/-/issues/18378). We might have to settle for heuristics here.

RyanGlScott:
This PR contains every commit that I needed to make `th-desugar` compile with `template-haskell-2.17.0.0`. In addition to the existing "Support explicit specificity" commit from #142, this includes two other commit:

* Deal with linear types–related changes in TH (which fixes #143)
* Add support for QualifiedDo 

Refer to each commit for an overview of what it does. Since each commit is relatively small, I decided to lump them into a single PR.

Subsumes #142.

### DIFF
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -15,6 +15,7 @@ module Language.Haskell.TH.Desugar.Core where
 import Prelude hiding (mapM, foldl, foldr, all, elem, exp, concatMap, and)
 
 import Language.Haskell.TH hiding (match, clause, cxt)
+import Language.Haskell.TH.Datatype.TyVarBndr
 import Language.Haskell.TH.Syntax hiding (lift)
 
 #if __GLASGOW_HASKELL__ < 709
@@ -105,7 +106,11 @@ dsExp (CaseE exp matches) = do
   matches' <- dsMatches scrutinee matches
   return $ DLetE [DValD (DVarP scrutinee) exp'] $
            DCaseE (DVarE scrutinee) matches'
-dsExp (DoE stmts) = dsDoStmts stmts
+dsExp (DoE
+#if __GLASGOW_HASKELL__ >= 900
+           _
+#endif
+           stmts) = dsDoStmts stmts
 dsExp (CompE stmts) = dsComp stmts
 dsExp (ArithSeqE (FromR exp)) = DAppE (DVarE 'enumFrom) <$> dsExp exp
 dsExp (ArithSeqE (FromThenR exp1 exp2)) =
@@ -691,9 +696,9 @@ dsDec (NewtypeD cxt n tvbs con derivings) =
   dsDataDec Newtype cxt n tvbs Nothing [con] derivings
 #endif
 dsDec (TySynD n tvbs ty) =
-  (:[]) <$> (DTySynD n <$> mapM dsTvb tvbs <*> dsType ty)
+  (:[]) <$> (DTySynD n <$> mapM dsTvbUnit tvbs <*> dsType ty)
 dsDec (ClassD cxt n tvbs fds decs) =
-  (:[]) <$> (DClassD <$> dsCxt cxt <*> pure n <*> mapM dsTvb tvbs
+  (:[]) <$> (DClassD <$> dsCxt cxt <*> pure n <*> mapM dsTvbUnit tvbs
                      <*> pure fds <*> dsDecs decs)
 #if __GLASGOW_HASKELL__ >= 711
 dsDec (InstanceD over cxt ty decs) =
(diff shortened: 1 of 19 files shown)


Language/Haskell/TH/Desugar/Reify.hs
```diff
@@ -1069,20 +1075,18 @@ build_kind arg_kinds res_kind =
     forall_vis _   _    = Nothing
 #endif
 
-tvb_is_kinded :: TyVarBndr -> Bool
+tvb_is_kinded :: TyVarBndr_ flag -> Bool
```
goldfirere, line 1078: What is this `TyVarBndr_` of which you speak? I can find its definition nowhere.

RyanGlScott: This is a compatibility shim that I defined in https://github.com/glguy/th-abstraction/pull/83 to make it possible to talk about `TyVarBndr`s in a backwards-compatible way now that `template-haskell-2.17.0.0` parameterizes `TyVarBndr`s by `flag`s. The gist of it is:

```hs
#if MIN_VERSION_template_haskell(2,17,0)
type TyVarBndr_ flag = TyVarBndr flag
#else
type TyVarBndr_ flag = TyVarBndr -- the flag is a phantom type
#endif
```

(This is the reason why I temporarily vendored in an in-flight version of `th-abstraction`, as it's not yet available on Hackage.)

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1588,23 +1681,33 @@ toposortTyVarsOf tys =
...
-dtvbName (DKindedTV n _) = n
+dtvbName :: DTyVarBndr flag -> Name
+dtvbName (DPlainTV n _)    = n
+dtvbName (DKindedTV n _ _) = n
+
+-- @mk_qual_do_name mb_mod orig_name@ will simply return @orig_name@ if
+-- @mb_mod@ is Nothing. If @mb_mod@ is @Just mod_@, then a new 'Name' will be
+-- returned that uses @mod_@ as the new module prefix. This is useful for
+-- emulating the behavior of the @QualifiedDo@ extension, which adds module
+-- prefixes to functions such as 'return' and ('>>=').
```
goldfirere, line 1692: Nitpick: `QualifiedDo` never affects `return`.

Language/Haskell/TH/Desugar/Subst.hs
```diff
@@ -62,22 +62,32 @@ substTy _ ty@DArrowT    = return ty
 substTy _ ty@(DLitT _)  = return ty
 substTy _ ty@DWildCardT = return ty
 
-substTyVarBndrs :: Quasi q => DSubst -> [DTyVarBndr]
-                -> (DSubst -> [DTyVarBndr] -> q a)
+substForallTelescope :: Quasi q => DSubst -> DForallTelescope
```
goldfirere, line 65: Why is this stuff CPS'd? Not new here, but I see no reason for it.

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: Good catch. I suppose I had `RecursiveDo` on the brain, since `RecursiveDo` can generate code involving `mfix` and `return` behind the scenes. `th-desugar` doesn't support `RecursiveDo`, however, so there's never any way for qualified `return`s to be created via desugaring.

I've updated the documentation to mention `(>>)` rather than `return`.

Language/Haskell/TH/Desugar/Subst.hs
RyanGlScott: Indeed, I only CPS'd `substForallTelescope` because `substTyVarBndrs` was already CPS'd by the time I got here. We could certainly remove the CPS'ing (at the expense of a breaking API change, but eh), but I'll save that change for #142, on which this PR is built.

PR #147, opened 2020-09-12
Make pragmaToTH error on unsupported DPragmas

### LINKED ISSUE #146: Inconsistent treatment of unsupported language features in L.H.TH.Desugar.Sweeten
In most parts of `th-desugar`, any code that deals with an unsupported language feature will simply throw an error. Here is an example from `L.H.TH.Desugar.Core`:

https://github.com/goldfirere/th-desugar/blob/67750a01d57939b136100310a74f60501fcf783c/Language/Haskell/TH/Desugar/Core.hs#L609-L610

And from `L.H.TH.Desugar.Sweeten`:

https://github.com/goldfirere/th-desugar/blob/67750a01d57939b136100310a74f60501fcf783c/Language/Haskell/TH/Desugar/Sweeten.hs#L59-L61

However, this convention is not applied uniformly in `L.H.TH.Desugar.Sweeten`. There is one particular function, `pragmaToTH`, that returns a `Maybe` rather than erroring out on unsupported language features:

https://github.com/goldfirere/th-desugar/blob/67750a01d57939b136100310a74f60501fcf783c/Language/Haskell/TH/Desugar/Sweeten.hs#L380-L384

This has a ripple effect on several other type signatures in `L.H.TH.Desugar.Sweeten`. For instance, the type of `decToTH` is `DDec -> [Dec]` instead the more sensible `DDec -> Dec`.

I propose that we change `pragmaToTH` to error out on unsupported pragmas, which would allow simplifying its type to `DPragma -> Pragma`, enabling other simplifications elsewhere in the module.

RyanGlScott:
Fixes #146.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -36,6 +36,24 @@ Version 1.12 [????.??.??]
   especially in the presence of GADT constructors. Moreover, this function was
   used in the implementation of `getRecordSelectors` function, so bugs in
   `conExistentialTvbs` often affected the results of `getRecordSelectors`.
+* The types of `decToTH`, `letDecToTH`, and `pragmaToTH` have changed:
+
+  ```diff
+  -decToTH :: DDec -> [Dec]
+  +decToTH :: DDec -> Dec
+
+  -letDecToTH :: DLetDec -> Maybe Dec
+  +letDecToTH :: DLetDec -> Dec
+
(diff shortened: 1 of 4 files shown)


goldfirere (approved):
If you think this new behavior is an improvement, I won't stand in your way. Thanks.

PR #150, opened 2020-12-21
Simplify the desugaring of lambda expressions

### LINKED ISSUE #149: Desugared lambda expressions with @-patterns are needlessly complicated
If you desugar the expression `\x@y -> f x y` with `th-desugar`, you'll get this:

```
λ> putStrLn $([e| \x@y -> f x y |] >>= dsExp >>= stringE . ppShow)
DLamE
  [ arg_6989586621679031072_6989586621679031073 ]
  (DCaseE
     (DVarE arg_6989586621679031072_6989586621679031073)
     [ DMatch
         (DVarP y_6989586621679031071)
         (DLetE
            [ DValD (DVarP x_6989586621679031070) (DVarE y_6989586621679031071)
            ]
            (DAppE
               (DAppE (DVarE f) (DVarE x_6989586621679031070))
               (DVarE y_6989586621679031071)))
     ])
```

Cleaned up a little, that's:

```hs
\arg -> case arg of { y -> let x = y in f x y }
```

Yuck. Why does `th-desugar` needlessly introduce that `arg` variable? We could just as well do this:

```hs
\y -> let x = y in f x y
```

I originally suspected that the way that `th-desugar`'s code is architected would make it difficult to achieve this, but upon a closer look, this is actually quite feasible. In fact, the resulting code is considerably cleaner than the original code. Patch incoming.

RyanGlScott:
`th-desugar`'s previous implementation was needlessly confusing, as there were no fewer than three functions that desugared lambda expressions in some way. This patch rips out two of those functions to leave `mkDLamEFromDPats` as the sole entrypoint for desugaring lambda expressions. This has some knock-on benefits:

* `mkDLamEFromDPats` now accepts a more general `Quasi` context rather than `DsMonad`.
* `mkDLamEFromDPats` now generates more direct code for certain lambda expressions that use `@`-patterns. For example, `\x@y -> f x y` would previously desugar to `\arg -> case arg of { y -> let x = y in f x y }`, but it now desugars to `\y -> let x = y in f x y`. This fixes #149.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -104,6 +104,11 @@ Version 1.12 [????.??.??]
   -substTyVarBndrs :: Quasi q => DSubst -> [DTyVarBndr flag] -> (DSubst -> [DTyVarBndr flag] -> q a) -> q a
   +substTyVarBndrs :: Quasi q => DSubst -> [DTyVarBndr flag] -> q (DSubst, [DTyVarBndr flag])
   ```
+* `mkDLamEFromDPats` has now generates slightly more direct code for certain
+  lambda expressions with `@`-patterns. For example, `\x@y -> f x y` would
+  previously desugar to `\arg -> case arg of { y -> let x = y in f x y }`, but
+  it now desugars to `\y -> let x = y in f x y`.
+* `mkDLamEFromDPats` now requires only a `Quasi` context instead of `DsMonad`.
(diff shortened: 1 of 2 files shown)


goldfirere (approved):
Yes, that's much better. Thanks!

Issue #148, opened 2020-12-18
GHC 9.2 checklist


RyanGlScott:
This issue serves as a reminder to do certain things before we ship a new major release that supports GHC 9.2:

* [x] Adapt to [type applications in patterns](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/2464)
* [x] Adapt to `OverloadedRecordDot`
* [x] Figure out what to do with #151
* [x] Fix an `-Wincomplete-record-updates` warning in `Test.Run`:

  ```
  [6 of 6] Compiling Main             ( Test/Run.hs, /home/ryanglscott/Documents/Hacking/Haskell/ci-maintenance/checkout/goldfirere/th-desugar/dist-newstyle/build/x86_64-linux/ghc-9.1.20210322/th-desugar-1.13/t/spec/build/spec/spec-tmp/Main.o, /home/ryanglscott/Documents/Hacking/Haskell/ci-maintenance/checkout/goldfirere/th-desugar/dist-newstyle/build/x86_64-linux/ghc-9.1.20210322/th-desugar-1.13/t/spec/build/spec/spec-tmp/Main.dyn_o )
  
  Test/Run.hs:98:31: warning: [-Wincomplete-record-updates]
      Pattern match(es) are non-exhaustive
      In a record-update construct:
          Patterns of type ‘Record’ not matched: MkRecord2 _ _
     |
  98 |              , "record"   ~: $test14_record   @=? $(dsSplice test14_record)
     |                               ^^^^^^^^^^^^^
  ```

  (**EDIT**: Worked around in 4b02b029ef9699c4bef012441a406f7993874355.)

RyanGlScott:
[`th-desguar-1.13`](https://hackage.haskell.org/package/th-desugar-1.13)

expipiplus1:
Thanks, @RyanGlScott!

PR #152, opened 2021-03-13
Support VTA in data constructor patterns

RyanGlScott:
The change at the heart of this patch involves `DConP`:

```diff
 data DPat
   = ...
-  | DConP Name         [DPat] -- fun (Just    x) = ...
+  | DConP Name [DType] [DPat] -- fun (Just @t x) = ...
   | ...
```

All other changes in this patch are a result of this.

Since this is a breaking change, I took the opportunity to bump the major version number to 1.13.

Addresses one bullet point of #148.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,21 @@
 `th-desugar` release notes
 ==========================
 
+Version 1.13 [????.??.??]
+-------------------------
+* Support GHC 9.2.
+* Add support for visible type application in data constructor patterns. As a
+  result of these changes, the `DConP` constructor now has an extra field to
+  represent type arguments:
+
+  ```diff
+   data DPat
+     = ...
+  -  | DConP Name         [DPat] -- fun (Just    x) = ...
+  +  | DConP Name [DType] [DPat] -- fun (Just @t x) = ...
+     | ...
+  ```
+
 Version 1.12 [2021.03.12]
 -------------------------
 * Support GHC 9.0.
--- a/Language/Haskell/TH/Desugar.hs
+++ b/Language/Haskell/TH/Desugar.hs
@@ -240,7 +240,7 @@ flattenDValD (DValD pat exp) = do
         DVarP n
           | n == name -> DVarP y
           | otherwise -> DWildP
-        DConP con ps -> DConP con (map (wildify name y) ps)
+        DConP con ts ps -> DConP con ts (map (wildify name y) ps)
         DTildeP pa -> DTildeP (wildify name y pa)
         DBangP pa -> DBangP (wildify name y pa)
         DSigP pa ty -> DSigP (wildify name y pa) ty
@@ -297,7 +297,7 @@ getRecordSelectors cons = merge_let_decs `fmap` concatMapM get_record_sels cons
             return $ concat
               [ [ DSigD name $ DForallT (DForallInvis con_tvbs)
                              $ DArrowT `DAppT` con_ret_ty `DAppT` field_ty
-                , DFunD name [DClause [DConP con_name
+                , DFunD name [DClause [DConP con_name []
                                          (mk_field_pats n (length fields) varName)]
                                       (DVarE varName)] ]
               | ((name, _strict, field_ty), n) <- zip fields [0..]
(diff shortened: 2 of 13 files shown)


goldfirere:
There is some wiggle room for design here.

## Design A

What you have written.

## Design B

Separate out universals from existentials:

```hs
| DConP Name [DType] [Maybe Name] [DPat]   -- universal patterns are DType, existential ones are always Names or blanks
```

Consumers of the new information will likely need to treat universals differently from existentials, and so this Design B enables easier consumption.

## Design C

```hs
data DConPatArg = PatArg DPat | UniversalArg DType | ExistentialArg (Maybe Name)

data DPat = ... | DConP Name [DConPatArg]
```

This version allows an arbitrary ordering of arguments. This extra flexibility is not needed today, but it will be needed tomorrow, and choosing this design eases migrations for our users.

------------------------

There is one thorny corner:

```hs
data Annoying where
  MkA :: (a ~ Int) => a -> Annoying
```

In `MkA`, the type variable `a` is existentially bound. Yet the pattern `MkA @Int x` is allowed, because `a ~ Int` is implied by the context. (See the Type Variables in Patterns paper.) This construction is bizarre, though. Furthermore, I think this case means that we can just skip the pattern if it's not a variable, treating the pattern above as `MkA @_ x`, which fits nicely in our story.

Dark corner of the dark corner:

```hs
data Irking where
  MkI :: forall a b. (a ~ Maybe b) => a -> Irking
```

Now, consider the pattern `MkI @(Maybe c) x`. This binds the type variable `c`, and so it's not equivalent to `MkI @_ x`. However, it *is* equivalent to `MkI @_ @c x`. But I think that will be hard to figure out, in general. So we either don't use `Maybe Name`s (just because of this! terrible!), or we warn in this case and blank out any non-type-variable existential type patterns, or we become clever and transform `MkI @(Maybe c) x` to `MkI @_ @c x`. (I *do* think such a transformation is always possible, but may be arbitrarily hard to compute, in case the equality constraint is, say, in a superclass of the actual user-written constraint.) I think using `Maybe Name` and warning on `Irking` is a nice spot to land.

This wrinkle, of course, goes away if we keep Design A. But I'd prefer the other designs, as they make desugared syntax easier to consume.

goldfirere:
I should add: I don't feel very strongly, especially since I'm barking orders from the armchair here (meaning: I lose some authority on this package, given that I'm not doing active maintenance.) But I just wanted to open the door to other approaches.

RyanGlScott:
> I should add: I don't feel very strongly, especially since I'm barking orders from the armchair here (meaning: I lose some authority on this package, given that I'm not doing active maintenance.)

I greatly value your feedback, so don't view it as barking orders at all. I'll admit that I hadn't really thought over alternative designs, so this is useful to consider.

Both Design B and Design C rely on the ability to distinguish universally and existentially quantified type variables, correct? It's worth noting that doing this has been somewhat of a pain point in the past, and previous versions of `th-desugar` used unreliable heuristics to compute them. In the end, they were unreliable enough that I removed functionality to compute existential type variables in 5eef50aad9a7db9ad284a427b791891398304fa8.

This isn't to say that I'm completely opposed to the idea of bringing this back, but if we do, we would need to come up with a story for how to conservatively guess the existential type variables. As I see it, one of the advantages of Design A is that we don't have to worry about the universal/existential distinction at all. Moreover, I don't have any particular use case for distinguishing between these sorts of type variables, but perhaps you have one in mind.

-----

To respond to your point about Design C allowing an arbitrary order of arguments: this is a good thought. It's tempting to try and anticipate what the syntax of data constructor patterns will be in a future version of GHC, and it seems quite likely that future GHCs will allow things like `f (MkT @a b @c d) = ...`. If we're in the business of looking into the future, however, that does make the current design of `DCon` insufficient:

https://github.com/goldfirere/th-desugar/blob/d6222162af1f1f8457e622635f422a94ca966091/Language/Haskell/TH/Desugar/AST.hs#L188-L189

This assumes that all of the type variables in a data constructor's type will be at the front, which wouldn't be the case in a world where `f (MkT @a b @c d) = ...` is possible. So I suppose that if we wanted to complete this thought experiment, we'd need to give `DCon` a new coat of paint as well.

In the end, I'm somewhat on the fence. I can see the appeal in trying to limit the number of times we have to make breaking changes in the future. At the same time, `th-desugar` has a funny habit of needing breaking changes with each new GHC release, so perhaps it's simply better to wait for GHC to change before `th-desugar` does.

goldfirere:
Perhaps I'm wrong about consumers. I can't think of a good example either. Given your points about the difficulty of telling universals from existentials (and your good point about inevitable breakage), I agree with your Design A.

goldfirere (approved):

PR #160, opened 2022-04-28
Desugar non-exhaustive case expressions using `mkErrorMatchExpr`

### LINKED ISSUE #159: Inconsistent treatment of non-exhausive guards in function definitions vs. `case` expressions
Consider this example:

```hs
{-# LANGUAGE EmptyCase #-}
{-# OPTIONS_GHC -ddump-splices #-}
module Bug where

import Language.Haskell.TH

$(do decs <- [d| f1, f2 :: () -> ()
                 f1 x | False = ()
                 f2 x = case x of y | False -> ()
               |]
     decs' <- dsDecs decs
     pure (sweeten decs'))

runF1, runF2 :: ()
runF1 = f1 (let x = x in x)
runF2 = f2 (let x = x in x)
```

This defines two functions, `f1` and `f2`, neither of which are exhaustive. The `runF1` and `runF2` functions invoke `f1` and `f2`, respectively, on a non-terminating argument.

What would you expect the outcomes of invoking `runF1` and `runF2` to be? Invoking `runF1` throws a `Non-exhaustive patterns` exception:

```
λ> runF1
*** Exception: Non-exhaustive patterns in f1_6989586621679031916
CallStack (from HasCallStack):
  error, called at Bug.hs:9:2 in fake-package-0-inplace:Bug
```

Invoking `runF2`, however, doesn't throw an exception at all. Instead, it forces the argument, causing `runF2` to loop forever at runtime! The `-ddump-splices` output reveals why this happens:

```
Bug.hs:(9,2)-(14,26): Splicing declarations
    do decs_a5IW <- [d| f1_a5IR, f2_a5IS :: () -> ()
                        f1_a5IR x_a5IT | False = ()
                        f2_a5IS x_a5IU = case x_a5IU of { y_a5IV | False -> () } |]
       decs'_a5IX <- dsDecs decs_a5IW
       pure (sweeten decs'_a5IX)
  ======>
    f1_a5KQ :: () -> ()
    f2_a5KP :: () -> ()
    f1_a5KQ arg_6989586621679031920_a5KV
      = case arg_6989586621679031920_a5KV of {
          x_a5KR
            -> case False of
                 True -> ()
                 False
                   -> error "Non-exhaustive patterns in f1_6989586621679031916" }
    f2_a5KP x_a5KS
      = case x_a5KS of {
          y_a5KT
            -> case False of
                 True -> ()
                 False -> case x_a5KS of }
```

While `f1` desugars its non-exhaustive guards to a final `Non-exhaustive patterns` error, `f2` desugars its non-exhaustive guards to an empty `case` expression. Technically speaking, throwing an exception is indistinguishable from non-termination in pure code, so it's hard to argue that desugaring changed `f2`'s semantics. It does, however, feel inconsistent to treat these cases differently, especially since if you were to define `f2` without desugaring, then `runF2` would also throw a `Non-exhaustive patterns` error instead of looping forever.

The reason for this discrepancy lies in the code that desugars `Clause`s (as in function definitions):

https://github.com/goldfirere/th-desugar/blob/2e27b31cfb64e5e43c12dde8ee29c3c8367bb134/Language/Haskell/TH/Desugar/Core.hs#L1160-L1164

Versus the code that desugars `Match`es (as in `case` expressions):

https://github.com/goldfirere/th-desugar/blob/2e27b31cfb64e5e43c12dde8ee29c3c8367bb134/Language/Haskell/TH/Desugar/Core.hs#L348-L351

I don't think there's a good reason for this discrepancy to exist, especially since it would be straightforward to change the latter code to use `maybeDCaseE "<error message>"` like the former code does.

RyanGlScott:
This makes desugaring non-exhaustive `case` expressions consistent with `th-desugar`'s treatment of all other forms of non-exhaustive expressions.

Fixes #159.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,13 @@
 `th-desugar` release notes
 ==========================
 
+Version next [????.??.??]
+-------------------------
+* Fix an inconsistency which caused non-exhaustive `case` expressions to be
+  desugared into uses of `EmptyCase`. Non-exhaustive `case` expressions are now
+  desugared into code that throws a "`Non-exhaustive patterns in...`" error at
+  runtime, just as all other forms of non-exhaustive expressions are desugared.
+
 Version 1.13 [2021.10.30]
 -------------------------
 * Support GHC 9.2.
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -346,7 +346,7 @@ dsMatches scr = go
     go [] = return []
     go (Match pat body where_decs : rest) = do
       rest' <- go rest
-      let failure = DCaseE (DVarE scr) rest'  -- this might be an empty case.
+      let failure = maybeDCaseE CaseAlt (DVarE scr) rest'
       exp' <- dsBody body where_decs failure
       (pat', exp'') <- dsPatOverExp pat exp'
       uni_pattern <- isUniversalPattern pat' -- incomplete attempt at #6
@@ -1181,6 +1181,8 @@ data MatchContext
     -- ^ A record update
   | MultiWayIfAlt
     -- ^ Guards in a multi-way if alternative
+  | CaseAlt
+    -- ^ Patterns and guards in a case alternative
 
 -- | Construct an expression that throws an error when encountering a pattern
 -- at runtime that is not covered by pattern matching.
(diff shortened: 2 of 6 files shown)


CHANGES.md
```diff
@@ -1,6 +1,13 @@
 `th-desugar` release notes
 ==========================
 
+Version next [????.??.??]
+-------------------------
+* Fix an inconsistency which caused non-exhaustive `case` expressions to be
+  desugared into uses of `EmptyCase`. Non-exhaustive `case` expressions are now
+  desugared into code that throws a "`Non-exhaustive patterns in...`" error at
+  runtime, just as all other forms of non-exhause expressions are desugared.
```
goldfirere, line 9:
```suggestion
  runtime, just as all other forms of non-exhaustive expressions are desugared.
```

goldfirere (approved):
Looks good. Thanks!

PR #162, opened 2022-04-30
Don't apply type-variable hack when expanding closed type families

### LINKED ISSUE #161: Expansion becomes stuck on closed type families with type variable arguments
Consider this example:

```hs
{-# LANGUAGE TypeFamilies #-}
module Main where

import Language.Haskell.TH

type family T1 a
type instance T1 (Maybe a) = a

type family T2 a where
  T2 (Maybe a) = a

$(pure [])

main :: IO ()
main = do
  putStrLn $(do t <- expandType (DConT ''T1 `DAppT` (DConT ''Maybe `DAppT` DVarT (mkName "a")))
                stringE $ show t)
  putStrLn $(do t <- expandType (DConT ''T2 `DAppT` (DConT ''Maybe `DAppT` DVarT (mkName "a")))
                stringE $ show t)
```

Here, we define two families `T1` and `T2` that are nearly identical, save for the fact that `T1` is an open type family and `T2` is a closed one. In `main`, we use `expandType` to expand the types `T1 (Maybe a)` and `T2 (Maybe a)` and print the results, which are:

```
λ> main
DVarT a
DAppT (DConT Main.T2) (DAppT (DConT GHC.Maybe.Maybe) (DVarT a))
```

Oddly enough, only `T1 (Maybe a)` expands, as `T2 (Maybe a)` is stuck. The reason for this discrepancy can be found in the code in `L.H.TH.Desugar.Expand`. Compare the case for open type families:

https://github.com/goldfirere/th-desugar/blob/12fa8373d65c263ca8dfe17a80427c6e7eeb2253/Language/Haskell/TH/Desugar/Expand.hs#L135-L140

To the case for closed type families:

https://github.com/goldfirere/th-desugar/blob/12fa8373d65c263ca8dfe17a80427c6e7eeb2253/Language/Haskell/TH/Desugar/Expand.hs#L160-L163

Where `args_ok` is:

https://github.com/goldfirere/th-desugar/blob/12fa8373d65c263ca8dfe17a80427c6e7eeb2253/Language/Haskell/TH/Desugar/Expand.hs#L125

On modern GHCs, the closed type family case requires that none of the argument types contain any type variables, but this restriction does not apply to open type families. Commit f3f23bec71db8b13544773d0190fc028c1c716d2 removed this restriction for open type families but kept the restriction for closed ones. I believe this is an oversight, and if I remove the `args_ok` check for closed type families, then `T2 (Maybe a)` expands to `a` as expected.

RyanGlScott:
`th-desugar`'s type expansion machinery contains a workaround for https://gitlab.haskell.org/ghc/ghc/-/issues/9262 that prevents `reifyInstances` from failing when the argument types contain type variables. This workaround was applied to both open and closed type families alike. While open type families certainly need the workaround, as their instances must be queried using `reifyInstances`, the workaround isn't necessary at all for closed type families, where all of the equations are immediately accessible.

This patch removes the workaround for closed type families and improves the documentation around the workaround to make it more obvious why it exists.

Fixes #161.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -7,6 +7,8 @@ Version next [????.??.??]
   desugared into uses of `EmptyCase`. Non-exhaustive `case` expressions are now
   desugared into code that throws a "`Non-exhaustive patterns in...`" error at
   runtime, just as all other forms of non-exhaustive expressions are desugared.
+* Fix a bug in which `expandType` would not expand closed type families when
+  applied to arguments containing type variables.
 
 Version 1.13 [2021.10.30]
 -------------------------
--- a/Language/Haskell/TH/Desugar/Expand.hs
+++ b/Language/Haskell/TH/Desugar/Expand.hs
@@ -30,9 +30,9 @@ module Language.Haskell.TH.Desugar.Expand (
   ) where
 
 import qualified Data.Map as M
-import Control.Monad
 #if __GLASGOW_HASKELL__ < 709
 import Control.Applicative
+import Control.Monad
 #endif
 import Language.Haskell.TH hiding (cxt)
 import Language.Haskell.TH.Syntax ( Quasi(..) )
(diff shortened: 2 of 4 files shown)


goldfirere (approved):
I don't see how the change to the code (making some code conditional on a GHC version) achieves the goal stated in the description (having the workaround apply only to open type families), but I believe you. So this is a pretty superficial code review -- do let me know if you want me to actually get into the details and understand this before approving.

RyanGlScott:
Ah, I should clarify the commit message a bit then. The part of this patch that actually fixes the bug is deleting the `args_ok` check in the `DClosedTypeFamilyD` case. All other changes involving CPP are just to make sure that unused definition warnings aren't generated on modern GHCs, since `args_ok`'s only other use is in the `DOpenTypeFamilyD` case, guarded by CPP.

PR #163, opened 2022-04-30
Require GHC 8.0 or later

RyanGlScott:
This requires GHC 8.0 as the minimum, removing support for GHC 7.8 and 7.10 in the process. As a consequence, we can now remove enormous chunks of unneeded CPP.

Now that we no longer use `template-haskell`'s `Strict` type, the `strictToBang` function no longer serves a useful purpose. I've opted to remove it, which makes this commit a breaking change (albeit a pretty obscure one).

While I was in town, I spruced up the code in various places to make use of more modern GHC features that previously weren't avaiable on all supported versions:

* We no longer need to explicitly derive `Typeable` instances, as GHC now does this automatically for all data types.
* We no longer need to bluntly disable `-Wstar-is-type`, as we can simply import `Data.Kind` unconditionally now.
* We now favor `-W`-style warnings over `-fwarn-`style ones, the former being the modern warning syntax in GHC.
* We now use `TypeApplications` to simplify some boilerplate in the `OMap` and `OSet` implementations.
* We now favor the `TemplateHaskellQuotes` extension over `TemplateHaskell`, as we do not need the full power of `TemplateHaskell` in most places.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -68,16 +68,6 @@ jobs:
             compilerVersion: 8.0.2
             setup-method: hvr-ppa
             allow-failure: false
-          - compiler: ghc-7.10.3
-            compilerKind: ghc
-            compilerVersion: 7.10.3
-            setup-method: hvr-ppa
-            allow-failure: false
-          - compiler: ghc-7.8.4
-            compilerKind: ghc
-            compilerVersion: 7.8.4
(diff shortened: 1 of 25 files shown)


goldfirere (approved):
LGTM, though I started scrolling more quickly once I got the general gist of what's going on. I'm grateful for CI.

RyanGlScott:
> I'm grateful for CI.

Hear hear.

PR #166, opened 2022-05-14
Use `withDict` to implement `bindIP` on recent GHCs

### LINKED ISSUE #165: Use `withDict` instead of `unsafeCoerce` to implement `bindIP`
Currently, `bindIP` is implemented using `unsafeCoerce`:

https://github.com/goldfirere/th-desugar/blob/d87ea5678dfd6726a0776db71468dbd97dda2985/Language/Haskell/TH/Desugar/Util.hs#L433-L434

Using GHC 9.4 or later, however, we can instead implement this using [`withDict`](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5573). This is safer to use, results in cleaner Core, and will be future-proof against future changes to "newtype classes" that are being considered [here](https://gitlab.haskell.org/ghc/ghc/-/issues/11715#note_428887).

See goldfirere/singletons#490 and ekmett/reflection#44 for related issues.

RyanGlScott:
Instead of `unsafeCoerce`, we can use the more recent `withDict` primitive from `GHC.Exts`. Among other benefits, this is more likely to be future-proof against possible changes to GHC's representation of "newtype classes" in the future.

Fixes #165.


RyanGlScott:
I decided to try and tackle `reflection`'s uses of `unsafeCoerce` after this, at which point I discovered that replacing `unsafeCoerce` with `withDict` broke `reflection`'s test suite. See [GHC#21575](https://gitlab.haskell.org/ghc/ghc/-/issues/21575). As a result, I'm going to mark this PR as a draft and wait for a resolution to that issue before proceeding.

RyanGlScott:
[GHC#21575](https://gitlab.haskell.org/ghc/ghc/-/issues/21575) is still not completely fixed as of GHC 9.4.1, so I think it's best to hold off on this.

Issue #157, opened 2021-11-10
GHC 9.4 checklist


RyanGlScott:
This issue serves as a reminder to do certain things before we ship a new major release that supports GHC 9.4:

* [ ] Support [`DefaultD`](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/5197).

RyanGlScott:
At this point, the only thing holding back a new release is the `th-lift-instances` dependency. It currently does not build with `containers-0.6.6` (bundled with GHC 9.4) due to bennofs/th-lift-instances#32.

ysangkok:
@RyanGlScott Is that dependency still needed? I see in the README that the minimum required version of GHC is 8.0.

Seems like the library is only needed with really old versions of GHC, as can be seen in phadej/file-embed-lzma#11.

This patch seems to make it compile with 9.4.1:

```diff
diff --git a/Language/Haskell/TH/Desugar/Lift.hs b/Language/Haskell/TH/Desugar/Lift.hs
index 50bb7e9..980bd24 100644
--- a/Language/Haskell/TH/Desugar/Lift.hs
+++ b/Language/Haskell/TH/Desugar/Lift.hs
@@ -20,7 +20,6 @@
 module Language.Haskell.TH.Desugar.Lift () where
 
 import Language.Haskell.TH.Desugar
-import Language.Haskell.TH.Instances ()
 import Language.Haskell.TH.Lift
 
 $(deriveLiftMany [ ''DExp, ''DPat, ''DType, ''DForallTelescope, ''DTyVarBndr
diff --git a/th-desugar.cabal b/th-desugar.cabal
index 234b8a7..59226fc 100644
--- a/th-desugar.cabal
+++ b/th-desugar.cabal
@@ -45,7 +45,7 @@ library
   build-depends:
       base >= 4.9 && < 5,
       ghc-prim,
-      template-haskell >= 2.11 && < 2.19,
+      template-haskell >= 2.11 && < 2.20,
       containers >= 0.5,
       mtl >= 2.1 && < 2.4,
       ordered-containers >= 0.2.2,
```

The tests don't compile though, since th-orphans depends on th-lift-instances.

RyanGlScott:
[...]

PR #169, opened 2022-05-31
Support TH constructs added in GHC 9.4

RyanGlScott:
This adds support for the following TH constructs, newly introduced in GHC 9.4:

* `PromotedInfixT` and `PromotedUInfixT`. These are desugared very similarly to `InfixT` and `UInfixT`.
* `DefaultD` (`default` declarations).
* `OpaqueP` (`OPAQUE` pragmas).
* `LamCasesE` (`\cases` expressions). These are desugared to ordinary lambda expressions, similarly to how `LamCaseE` is desugared.

For more details, refer to the individual commit messages.

This addresses four check boxes in #157.

### DIFF
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -1146,10 +1146,8 @@ dsType (AppT t1 t2) = DAppT <$> dsType t1 <*> dsType t2
 dsType (SigT ty ki) = DSigT <$> dsType ty <*> dsType ki
 dsType (VarT name) = return $ DVarT name
 dsType (ConT name) = return $ DConT name
-  -- the only difference between ConT and PromotedT is the name lookup. Here, we assume
-  -- that the TH quote mechanism figured out the right name. Note that lookupDataName name
-  -- does not necessarily work, because `name` has its original module attached, which
-  -- may not be in scope.
+-- The PromotedT case is identical to the ConT case above.
+-- See Note [Desugaring promoted types].
 dsType (PromotedT name) = return $ DConT name
 dsType (TupleT n) = return $ DConT (tupleTypeName n)
 dsType (UnboxedTupleT n) = return $ DConT (unboxedTupleTypeName n)
@@ -1180,6 +1178,8 @@ dsType (ForallVisT tvbs ty) =
   DForallT <$> (DForallVis <$> mapM dsTvbUnit tvbs) <*> dsType ty
 #endif
 #if __GLASGOW_HASKELL__ >= 903
+-- The PromotedInfixT case is identical to the InfixT case above.
+-- See Note [Desugaring promoted types].
 dsType (PromotedInfixT t1 n t2) = dsInfixT t1 n t2
 dsType PromotedUInfixT{} = dsUInfixT
 #endif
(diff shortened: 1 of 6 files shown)


Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1145,6 +1179,10 @@ dsType (ImplicitParamT n t) = do
 dsType (ForallVisT tvbs ty) =
   DForallT <$> (DForallVis <$> mapM dsTvbUnit tvbs) <*> dsType ty
 #endif
+#if __GLASGOW_HASKELL__ >= 903
+dsType (PromotedInfixT t1 n t2) = dsInfixT t1 n t2
```
goldfirere, line 1183: This doesn't look right. Suppose we have

```
data a :+: b = Inl a | Inr b
data Exp a = ... | Exp :+: Exp
```

I would expect `PromotedInfixT ty1 (mkName ":+:") ty2` to desugar to use the `:+:` from `Exp`, not the one from `:+:`. But it looks like this implementation treats `PromotedInfixT` and `InfixT` identically. C.f. the treatment in `GHC.ThToHs`, which handles the name differently.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1306,6 +1352,10 @@ dsPred t@(ForallVisT {}) =
 #if __GLASGOW_HASKELL__ >= 900
 dsPred MulArrowT = impossible "Linear arrow seen as head of constraint."
 #endif
+#if __GLASGOW_HASKELL__ >= 903
+dsPred (PromotedInfixT t1 n t2) = (:[]) <$> dsInfixT t1 n t2
```
goldfirere, line 1356: This looks wrong for similar reasons as my comment above (though maybe there is no correct usage of `PromotedInfixT` in this context).

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: This is a fair observation, but there is existing precedent for treating these identically. See this comment in between `ConT` and `PromotedT`, which are also desugared in the same way:

https://github.com/goldfirere/th-desugar/blob/e4151fec6a74fd806d5f3dbae2b6058d86dd8f10/Language/Haskell/TH/Desugar/Core.hs#L1114-L1119

Would you characterize this as being wrong as well? If so, it may be worth fixing this first before addressing the incongruity between `InfixT` and `PromotedInfixT`.

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: > though maybe there is no correct usage of `PromotedInfixT` in this context

Good point. I hadn't noticed that the `dsPred` case for `PromotedT` is marked as `impossible`, so for the sake of consistency, I've marked the `PromotedInfixT` as `impossible` as well.

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: I've added a `Note [Desugaring promoted types]` and referenced it from the `dsType` cases for `PromotedT` and `PromotedInfixT`.

goldfirere (approved):

PR #172, opened 2022-11-07
Local reification for pattern synonym record selectors

### LINKED ISSUE #137: Local reification misses pattern synonym field selectors
The following program, when run, prints `Nothing`:

```hs
{-# LANGUAGE PatternSynonyms #-}
module Main where

import Language.Haskell.TH

newtype Id a = MkId a

main :: IO ()
main = putStrLn
  $(do decs <- [d| pattern P :: a -> Id a
                   pattern P {unP} = MkId unP |]
       info <- withLocalDeclarations decs $ dsReify $ mkName "unP"
       stringE $ show info)
```

But it needn't do so. It should be possible to use local reification to cobble together a type signature for `unP`, much like we do for field selectors in data constructors.

RyanGlScott:
This augments `th-desugar`'s local reification machinery with the power to reconstruct the types of pattern synonym record selectors. See the Haddocks on the new `maybeReifyPatSynRecSelector` function for a description of how this works.

Fixes #137.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 `th-desugar` release notes
 ==========================
 
+next [????.??.??]
+-----------------
+* Local reification can now reify the types of pattern synonym record
+  selectors.
+
 Version 1.14 [2022.08.23]
 -------------------------
 * Support GHC 9.4.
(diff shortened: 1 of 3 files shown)


goldfirere (approved):
All looks good (after a somewhat in-depth look). Thanks!

PR #173, opened 2022-11-11
Make local reification aware of GADT type variable binder ordering

### LINKED ISSUE #171: Locally quantified GADT record selectors can quantify type variables in the wrong order
In this program:

```hs
#!/usr/bin/env cabal
{- cabal:
build-depends: base, pretty-show, template-haskell, th-desugar
-}

{-# LANGUAGE GADTs #-}
{-# LANGUAGE ScopedTypeVariables #-}
{-# LANGUAGE StandaloneKindSignatures #-}
module Main where

import Data.Kind (Type)
import Language.Haskell.TH hiding (Type)
import Text.Show.Pretty

type T1 :: Type -> Type -> Type
data T1 a b where
  MkT1 :: forall b a. { unT1 :: b } -> T1 a b

main :: IO ()
main = putStrLn
  $(do decs <- [d| type T2 :: Type -> Type -> Type
                   data T2 a b where
                     MkT2 :: forall b a. { unT2 :: b } -> T2 a b |]
       info <- withLocalDeclarations decs $ dsReify $ mkName "unT2"
       stringE $ ppShow info)
```

The type of `unT1` quantifies `b` before `a`:

```
> :set -fprint-explicit-foralls
> :t unT1
unT1 :: forall b a. T1 a b -> b
```

The `unT2` record selector should have the same type, but the type that `th-desugar` gives to it quantifies the type variables in the opposite order:

```
$ cabal run Bug.hs
Just
  (DVarI
     unT2
     (DForallT
        (DForallInvis
           [ DPlainTV a_6989586621679032956 SpecifiedSpec
           , DPlainTV b_6989586621679032955 SpecifiedSpec
           ])
        (DAppT
           (DAppT
              DArrowT
              (DAppT
                 (DAppT
                    (DConT T2_6989586621679032950) (DVarT a_6989586621679032956))
                 (DVarT b_6989586621679032955)))
           (DVarT b_6989586621679032955)))
     Nothing)
```

The culprit are these lines:

https://github.com/goldfirere/th-desugar/blob/b64e24b67c91991f5d9b97b055d05693fd99cb9b/Language/Haskell/TH/Desugar/Reify.hs#L330-L331

This will simply gather the free variables in `[T1 a b, b]` in left-to-right order. This is not the right thing to do in the event that a GADT constructor manually specifies a `forall`, however.

RyanGlScott:
This is implemented by remembering the explicit `forall` in a GADT constructor's type signature (if one exists) in `RecSelGADT` and using it later to construct the record selector's type signature.

Fixes #171.

### DIFF
--- a/Language/Haskell/TH/Desugar/Reify.hs
+++ b/Language/Haskell/TH/Desugar/Reify.hs
@@ -336,9 +336,17 @@ maybeReifyCon n _decs ty_name ty_args cons
           , sel_ty
           , h98_res_ty
           )
-        RecSelGADT sel_ty con_res_ty ->
-          ( changeTVFlags SpecifiedSpec $
-            freeVariablesWellScoped [con_res_ty, sel_ty]
+        RecSelGADT mb_con_tvbs sel_ty con_res_ty ->
+          let -- If the GADT constructor type signature explicitly quantifies
+              -- its type variables, make sure to use that same order in the
+              -- record selector's type.
+              con_tvbs' =
+                case mb_con_tvbs of
+                  Just con_tvbs -> con_tvbs
+                  Nothing ->
+                    changeTVFlags SpecifiedSpec $
+                    freeVariablesWellScoped [con_res_ty, sel_ty] in
+          ( con_tvbs'
           , sel_ty
           , con_res_ty
           )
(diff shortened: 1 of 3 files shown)


Language/Haskell/TH/Desugar/Reify.hs
```diff
@@ -336,9 +336,17 @@ maybeReifyCon n _decs ty_name ty_args cons
...
-        RecSelGADT sel_ty con_res_ty ->
-          ( changeTVFlags SpecifiedSpec $
-            freeVariablesWellScoped [con_res_ty, sel_ty]
+        RecSelGADT mb_con_tvbs sel_ty con_res_ty ->
+          let -- If the GADT constructor type signature explicitly quantifies
+              -- its type variables, make sure to use that same order in the
+              -- record selector's type.
+              con_tvbs' =
+                case mb_con_tvbs of
+                  Just con_tvbs -> con_tvbs
```
goldfirere, line 345: I think this list can be too long.

For example, if we have

```hs
data G a where
  MkG :: forall a b. { x :: a, y :: b } -> G a
```

then we have

```
ghci> :t x
x :: forall a. G a -> a
```

with `-fprint-explicit-foralls`. So I think this list needs to be thinned to contain only those variables free in the result or field. (Of course, if the latter isn't a subset of the former, we have other problems.) You could do this thinning either here or when the list is stored.

goldfirere (changes requested):

Language/Haskell/TH/Desugar/Reify.hs
RyanGlScott: Great catch, I had forgotten entirely about naughty record selectors. I have fixed this oversight and augmented the test case to ensure that this is accounted for.

goldfirere (approved):

Issue #174, opened 2023-01-13
Pattern guards are not a sufficient replacement for view patterns


expipiplus1:
The implementation for desugaring view patterns is `dsPat (ViewP _ _) = dsPat (ViewP _ _) =   fail "View patterns are not supported in th-desugar. Use pattern guards instead."`

It's not possible to use pattern guards in pattern synonyms, so desugaring the following will fail: 

```haskell
  x :: [DDec] <-
    desugar
      [ PatSynD
          (mkName "pat")
          (RecordPatSyn [])
          (ExplBidir [])
          ( ViewP
              (VarE (mkName "a"))
              (TupP [])
          )
      ]
```

And of course it's not possible to construct such pattern synonyms!

There's definitely a complexity/usefulness tradeoff to be struck; although I'd like to be able to use view patterns with th-desugar I would understand if this is closed with something like "not worth the complexity".

RyanGlScott:
This is most likely a consequence of two historical coincidences:

1. This code was first written before pattern synonyms were a thing. At the time, any use of view patterns could be straightforwardly converted to pattern guards (which then can be desugared to `case` expressions), so view patterns were excluded from `th-desugar`'s AST to keep things simpler.
2. `th-desugar` was originally forked from the `singletons` library. There is no reasonable way to make view patterns work in the context of `singletons`, so that was another reason to keep view patterns out of the `th-desugar` AST.

Now that pattern synonyms are a thing (and that we have `th-desugar` users besides `singletons`), we should revisit these assumptions. I agree that view patterns are quite essential for defining many kinds of pattern synonyms, and there really isn't a viable alternative to them—at least, not in this particular spot. Some possible ways to handle this:

1. Add a `DViewP :: DExp -> DPat -> DPat` data constructor to `DPat`. This would allow defining view patterns, but now all downstream consumers of `th-desugar` will be forced to reckon with view patterns in any place where `DPat`s can occur.
2. Leave `DPat` alone, and instead add a new `DPatSynPat` data type:

   ```hs
   data DPatSynPat
     = DPatSynViewPat DExp DPatSynViewPat
     | DPatSynPat DPat
   ``` 

[...]

PR #175, opened 2023-01-14
Support as-patterns in pattern synonyms with view patterns

expipiplus1:
I guess a bit of a WIP, Maybe needs neatening and tests

See https://github.com/goldfirere/th-desugar/issues/174


expipiplus1:
View patterns too, a tiny change on top of as-patterns.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -558,29 +561,71 @@ mk_tuple_dpat name_set =
...
+  dsAsP :: Name -> DPat -> m DPat
+  dsViewP :: DExp -> DPat -> m DPat
+
+instance DsMonad q => PatM (PatWithExp q) where
+  dsAsP :: Name -> DPat -> PatWithExp q DPat
+  dsAsP name pat = do
+    pat' <- lift $ removeWilds pat
+    tell [(name, dPatToDExp pat')]
+    return pat'
+  dsViewP _ _ = fail "View patterns are not supported in th-desugar. Use pattern guards instead."
```
RyanGlScott, line 615: This error message is somewhat misleading, as `th-desugar` now does support view patterns in limited circumstances.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -558,29 +561,71 @@ mk_tuple_dpat name_set =
...
+-- further patterns for case analysis on associated names. View patterns are
+-- also returned by pairing their expression and pattern
+newtype PatWithCases q a = PatWithCases { unPatWithCases :: WriterT [(DExp, DPat)] q a }
+  deriving newtype ( Functor, Applicative, Monad
+                   , MonadTrans, Fail.MonadFail, MonadIO
+                   , MonadWriter [(DExp, DPat)]
+                   , Quasi, DsMonad
+                   )
+
+class PatM m where
```
RyanGlScott, line 605: This class deserves some Haddocks.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -558,29 +561,71 @@ mk_tuple_dpat name_set =
...
+newtype PatWithExp q a = PatWithExp { unPatWithExp :: WriterT [(Name, DExp)] q a }
+  deriving newtype ( Functor, Applicative, Monad
+                   , MonadTrans, Fail.MonadFail, MonadIO
+                   , MonadWriter [(Name, DExp)]
+                   , Quasi, DsMonad
+                   )
+
+-- | Desugaring a pattern where any as-patterns are returned in reverse order as
+-- further patterns for case analysis on associated names. View patterns are
+-- also returned by pairing their expression and pattern
```
RyanGlScott, line 597: It would be worth stating that this _only_ used when desugaring the pattern on the RHS of a pattern synonym.

RyanGlScott (commented):
Thanks for the patch! I haven't had a chance to look at the finer details of how this works, but the general idea looks right.

Two possible ideas for making the generated code a bit more compact:

1. This:
   
   ```hs
   pattern P x <- (id -> x)
   ```
   
   Currently desugars to this:
   
   ```hs
   pattern P x_1 <- (\arg_2 -> case arg_2 of
                                   viewed_3 -> case GHC.Base.id viewed_3 of
                                                   x_1 -> x_1 -> x_1)
   ```
   
   The first `case` expression isn't necessary, as it just binds a simple `viewed_3`. Nor is `arg_2` used anywhere in the remainder of the expression. I think it should be possible to check if the very first pattern to be bound (`viewed_3`, in this example) is a bare `DVarP`, and if so, elide the intermediate `case` expression.


2. Regarding your "uncomfortable unpacking/repacking" comment at https://github.com/goldfirere/th-desugar/issues/174#issuecomment-1382674047, what if instead of desugaring this:
   
   ```hs
   data Foo a b = Foo a b
   
   pattern P x a b <- x@(Foo a b)
   ```
   
   To this (cleaned up slightly):
   
   ```hs
   pattern P x_1 a_2 b_3 <- ((\x_1 -> case x_1 of Foo a_2 b_3 -> (x_1, a_2, b_3)) -> (x_1, a_2, b_3))
   ```
   
   We instead desugared to something like this?
   
   ```hs
   pattern P x_1 a_2 b_3 <- ((\x_1 -> (x_1, x_1)) -> (x_1, Foo a_2 b_3))
   ```
   
   That is, if a pattern doesn't contain any view patterns or as-patterns as subpatterns, then match on it in the pattern to the right of the `->` instead of matching on it in the expression to the left of the `->`. This avoids having to unpack it in the expression on the left only to repack it again for the benefit of the pattern on the right.

I haven't thought deeply about how to implement these ideas, so I'm not sure if they work well with the current approach to desugaring patterns.

expipiplus1:
ok, I think I've addressed the comments, as well as added view pattern support everywhere (it fell out very naturally)

I've keps `dsPatX` around for a deprecation cycle if we care about that.

expipiplus1:
err, actually I think there's a bug here I still need to sort!

expipiplus1:
The bug here is that now `dsPatOverExp` return a less fallible pattern, having moved most of the matching on `as-patterns` into the subsequent let expression. The consequence of this is apparent in the following:

```
do 
  let xs = [1,2]
  x@1 <- xs
  pure x
```

This used to desugar to the equivalent of:

```  
do 
  let xs = [1,2]
  1 <- xs
  let x = 1
  pure x
 ```

But now it desugars to the equivalent of:

```
do
  let xs  = [1,2]
  x <- xs
  let 1 = x
  pure x
```

Apart from it being incorrect, I do prefer the second one as there isn't any pattern -> expression transmutation happening.

File: Language/Haskell/TH/Desugar/Core.hs [hunk 4]
```diff
@@ -558,29 +561,71 @@ mk_tuple_dpat name_set =
...
 -- | Desugar a pattern, returning a list of (Name, DExp) pairs of extra
 -- variables that must be bound within the scope of the pattern
 dsPatX :: DsMonad q => Pat -> q (DPat, [(Name, DExp)])
-dsPatX = runWriterT . dsPat
+dsPatX = runWriterT . unPatWithExp . dsPat
+
+-- | Desugar a pattern, returning a list of (Name, DExp) pairs of extra
+-- variables that must be bound within the scope of the pattern
+dsPatC :: DsMonad q => Pat -> q (DPat, [(DExp, DPat)])
+dsPatC = fmap (second reverse) . runWriterT . unPatWithCases . dsPat
```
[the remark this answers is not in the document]
expipiplus1: It was probably cases, I've changed the name to `dsPat'` as it's now the canonical way of desugaring patterns.

RyanGlScott:
OK. I could take a look, but there is currently a [build failure](https://github.com/goldfirere/th-desugar/actions/runs/3926783337/jobs/6712823624#step:18:35) in the patch, as caught by CI:

```
[ 9 of 14] Compiling Language.Haskell.TH.Desugar.Core ( Language/Haskell/TH/Desugar/Core.hs, /__w/th-desugar/th-desugar/dist-newstyle/build/x86_64-linux/ghc-9.2.2/th-desugar-1.14/build/Language/Haskell/TH/Desugar/Core.o, /__w/th-desugar/th-desugar/dist-newstyle/build/x86_64-linux/ghc-9.2.2/th-desugar-1.14/build/Language/Haskell/TH/Desugar/Core.dyn_o )

Language/Haskell/TH/Desugar/Core.hs:806:22: error:
    Variable not in scope:
      dsPatWithViewPatterns :: Pat -> q (DPat, t0 (DPat, DExp))
    |
806 |   (pat', subPats) <- dsPatWithViewPatterns pat
    |                      ^^^^^^^^^^^^^^^^^^^^^

Language/Haskell/TH/Desugar/Core.hs:950:19: error:
    Variable not in scope:
      dsPatWithViewPatterns :: Pat -> q (DPat, [(DPat, DExp)])
    |
950 |   (pat', vars) <- dsPatWithViewPatterns pat
    |                   ^^^^^^^^^^^^^^^^^^^^^
```

expipiplus1:
Ah, I didn't mean that I had uncovered an existing bug; I've introduced a new one. Before the "extra patterns" were designed/assumed to be infallible, but now they're not which leads to problems in do notation.

It shouldn't be the end of the world to fix, although it's tempting just to reinstate the old "no view patterns" behaviour.

RyanGlScott:
> It shouldn't be the end of the world to fix, although it's tempting just to reinstate the old "no view patterns" behaviour.

OK. For what it's worth, I would be fine with reinstating the old behavior, opening a separate issue to track the remaining tasks.

expipiplus1:
Ok, I reverted to the previous behaviour and opened https://github.com/goldfirere/th-desugar/issues/177

expipiplus1:
I also snuck in a deprecation notice on `dsPatX` hoping to encourage people to use the more versatile `dsPat'`.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -558,29 +561,89 @@ mk_tuple_dpat name_set =
 -- is the entire scope of the variables bound in the pattern.
 dsPatOverExp :: DsMonad q => Pat -> DExp -> q (DPat, DExp)
 dsPatOverExp pat exp = do
-  (pat', vars) <- runWriterT $ dsPat pat
+  (pat', vars) <- runPatWithExp $ dsPat pat
```
RyanGlScott, line 564: Is this equivalent to just calling `dsPatX` here?

```suggestion
  (pat', vars) <- dsPatX pat
```

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -876,9 +954,9 @@ dsLetDec (FunD name clauses) = do
   clauses' <- dsClauses (FunRhs name) clauses
   return ([DFunD name clauses'], id)
 dsLetDec (ValD pat body where_decs) = do
-  (pat', vars) <- dsPatX pat
+  (pat', vars) <- dsPat' pat
```
RyanGlScott, line 957: This changes the way that `let` bindings are desugared, which is more ambitious than the title of the PR would suggest. I haven't thought deeply about this change, but could this potentially run into issues similar to the ones documented in #177? Or is this change fine because `let` bindings don't treat fallible patterns the same way that `do`-notation does?

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -558,29 +561,89 @@ mk_tuple_dpat name_set =
...
--- variables that must be bound within the scope of the pattern
+-- variables that must be bound within the scope of the pattern. These
+-- correspond to any as-patterns in the original pattern.
+--
+-- View patterns are not supported, please use pattern guards instead or 'dsPat''
 dsPatX :: DsMonad q => Pat -> q (DPat, [(Name, DExp)])
-dsPatX = runWriterT . dsPat
+dsPatX = runPatWithExp . dsPat
+
+-- | Desugar a pattern, returning a list of (DExp, DPat) pairs of extra
```
RyanGlScott, line 583: The comment here says `(DExp, DPat)`, but the type uses `(DPat, DExp)`. I don't have a strong opinion about which order to use, but it would be good to pick a single convention and use it consistently.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -558,29 +561,89 @@ mk_tuple_dpat name_set =
...
-  (pat', vars) <- runWriterT $ dsPat pat
+  (pat', vars) <- runPatWithExp $ dsPat pat
   let name_decs = map (uncurry (DValD . DVarP)) vars
   return (pat', maybeDLetE name_decs exp)
 
 -- | Desugar multiple patterns. Like 'dsPatOverExp'.
 dsPatsOverExp :: DsMonad q => [Pat] -> DExp -> q ([DPat], DExp)
 dsPatsOverExp pats exp = do
-  (pats', vars) <- runWriterT $ mapM dsPat pats
+  (pats', vars) <- runPatWithExp $ mapM dsPat pats
```
RyanGlScott, line 571: This is _almost_ `dsPatX`, although the types don't quite match up. It might be worth leaving a comment to this effect just in case someone comes along later deciding to address #177, since this is a line of code that would need to be changed in order to accomplish that.

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -558,29 +561,89 @@ mk_tuple_dpat name_set =
...
+                   , MonadWriter [(DPat, DExp)]
+                   , Quasi, DsMonad
+                   )
+
+-- | The eliminator for 'PatWithCases'
+runPatWithCases :: PatWithCases q a -> q (a, [(DPat, DExp)])
+runPatWithCases = runWriterT . unPatWithCases
+
+-- | This class permits different implementations of matching on as-patterns
+-- and view-patterns in order to preserve the legacy behavior of 'dsPatX'.
```
RyanGlScott, line 621: Is it really legacy behavior? `dsPatX` is still being used, after all!

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -558,29 +561,89 @@ mk_tuple_dpat name_set =
...
+  dsViewP :: DExp -> DPat -> m DPat
+
+instance DsMonad q => PatM (PatWithExp q) where
+  dsAsP :: Name -> DPat -> PatWithExp q DPat
+  dsAsP name pat = do
+    pat' <- lift $ removeWilds pat
+    tell [(name, dPatToDExp pat')]
+    return pat'
+  dsViewP _ _ =
+    fail "PatWithExp and dsPatX don't support view-patterns, please use pattern guards instead."
```
RyanGlScott, line 633: I wonder if this error message should mention `PatWithExp` and `dsPatX` at all, given that most users will bump into this restriction without ever directly using either of those things. Perhaps it would be better to say `View patterns aren't supported in patterns in do-notation` instead?

Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -747,10 +810,25 @@ dsDec (RoleAnnotD n roles) = return [DRoleAnnotD n roles]
...
+    then pure (DPatSynPat pat')
+    else do
+      -- All the names we need to bind for the pattern
+      let allNames = allNamesIn args
+      -- RHS of the lambda
+      let r = mkTupleDExp (DVarE <$> allNames)
+      -- RHS of the view pattern
+      let p = mkTupleDPat (DVarP <$> allNames)
+      -- Fold up the sub-matches as successive cases
+      let c = foldr (\(asPat, asExp) e -> DCaseE (asExp) [DMatch asPat e])
```
RyanGlScott, line 824: Unnecessary parentheses:

```suggestion
      let c = foldr (\(asPat, asExp) e -> DCaseE asExp [DMatch asPat e])
```

RyanGlScott (commented):
Thank you so much for your continued work on this!

As one last suggestion, do you feel up to adding some test cases exercising this new functionality?

Language/Haskell/TH/Desugar/Core.hs
expipiplus1: Good catch, it does actually change the meaning, however neither are always correct!

I've opened https://github.com/goldfirere/th-desugar/issues/178

Language/Haskell/TH/Desugar/Core.hs
expipiplus1: Ah, the comment is wrong. I settled `DPat` on the left and `DExp` on the right as that's usually the way we see them in the source

Language/Haskell/TH/Desugar/Core.hs
expipiplus1: I'll keep the existing behaviour for now.

expipiplus1:
Thank you for taking the time for a thorough review @RyanGlScott!

For the tests, it would be really nice to be able to describe them like so

```haskell
test_t175 :: [Assertion]
test_t175 =
  fmap (uncurry (@=?))
  $(do
    Syn.lift =<< (traverse
      (\(a, e) -> (,) <$> (sweeten @_ @[DDec] <$> (desugar =<< a)) <*> e)
      [ ( [d| pattern P = (id -> ()) |]
        , [d| pattern P = (\x -> case id x of () -> () -> ()) |]
        )
      , ( [d| pattern P x <- x@3 |]
        , [d| pattern P x <- (\x -> case x of 3 -> x -> x) |]
        )
      , ([d| pattern P = ()|], [d| pattern P = ()|])
      ])
   )
```

But with something instead of `@=?` which checks up to alpha-equivalence. I couldn't find anything in th-desugar for this kind of test. Would this be something worth adding or is there a simpler test-setup you'd recommend?

RyanGlScott:
Most of the unit tests are located here:

https://github.com/goldfirere/th-desugar/blob/92c07bd88b74efae0d1b6b7bb5d52049723a009a/Test/Run.hs#L77-L154

Where each `test_*` function is located here:

https://github.com/goldfirere/th-desugar/blob/92c07bd88b74efae0d1b6b7bb5d52049723a009a/Test/Splices.hs#L136-L352

The problem with this approach is that it tests things by splicing in _expressions_. However, it's not entirely straightforward to define an expression containing a pattern synonym definition. You can't do something like `let { pattern P = (id -> ()) } in f P = 42`, as pattern synonyms must be defined at the top level.

One way we could accomplish this would be to define the pattern synonym definitions elsewhere, and then define expressions in terms of those pattern synonyms. We would need two copies of each pattern synonym: one using the original pattern synonym syntax, and another using the desugared syntax. We could define them in separate modules (similar to how we have both `Test/Decs.hs` and `Test/DsDecs.hs`) and import them qualified to avoid name clashes.

Testing alpha-equivalence would be another way to accomplish this, but nothing in `th-desugar` quite implements that. The closest that we have is the [`eqTH`](https://github.com/goldfirere/th-desugar/blob/92c07bd88b74efae0d1b6b7bb5d52049723a009a/Test/Splices.hs#L127-L128) function, which checks that two things are `Eq`ual to each other after removing the numeric suffixes from each type variable.

RyanGlScott:
How is this going, @expipiplus1? Do you need help with anything?

expipiplus1:
Other life things have gotten in the way a bit! I would like to finalize this, but it's not at the top of my plate at the moment, sorry.

I think there's also a bug, along the lines of: `(foo -> Left x)` being transformed to `(\a -> case foo a of Left x -> x)` (something like that, where fallible matches on the rhs of a view pattern were transformed into infallible (runtime pattern match error) case analysis in the lhs. So very annoyingly we need to add a catch all case to these and handle everything in `Maybe resultTuple` (or some unboxed variant to avoid any allocation).

RyanGlScott:
Thanks for the update! (I was worried that this was in a finished state and only being delayed due to figuring out how to add tests, in which case I could pick it up from there.)

PR #179, opened 2023-02-25
Support `type data` declarations

### LINKED ISSUE #170: GHC 9.6 checklist
This issue serves as a reminder to do certain things before we ship a new major release that supports GHC 9.6:

* [x] Support [`type data`](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9408)
* [x] Hackage release

RyanGlScott:
This patch:

* Adds a `TypeData` constructor to the `NewOrData` type to permit the possibility of `type data` declarations, which were introduced in GHC 9.6.
* Updates the surrounding code in `th-desugar` to desugar and sweeten `type data` declarations appropriately.
* Renames `NewOrData` to `NameFlavor`, as the old `NewOrData` name misleadingly suggests that there are only two possible options when there are now three.

Fixes #170.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,23 @@
 
 next [????.??.??]
 -----------------
+* Support GHC 9.6.
+* The `NewOrData` data type has been renamed to `DataFlavor` and extended to
+  support `type data` declarations:
+
+  ```diff
+  -data NewOrData  = NewType | Data
+  +data DataFlavor = NewType | Data | TypeData
+  ```
+
+  Desugaring upholds the following properties regarding `TypeData`:
+
+  * A `DDataD` with a `DataFlavor` of `TypeData` cannot have any deriving
+    clauses or datatype contexts, and the `DConFields` in each `DCon` will be a
+    `NormalC` where each `Bang` is equal to
+    `Bang NoSourceUnpackedness NoSourceStrictness`.
+  * A `DDataInstD` can have a `DataFlavor` of `NewType` or `Data`, but not
+    `TypeData`.
 * Local reification can now reify the types of pattern synonym record
   selectors.
 * Fix a bug in which the types of locally reified GADT record selectors would
--- a/Language/Haskell/TH/Desugar/AST.hs
+++ b/Language/Haskell/TH/Desugar/AST.hs
@@ -105,14 +105,28 @@ data DLetDec = DFunD Name [DClause]
              | DPragmaD DPragma
              deriving (Eq, Show, Data, Generic, Lift)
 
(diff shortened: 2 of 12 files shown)


CHANGES.md
```diff
@@ -3,6 +3,21 @@
...
+* Support GHC 9.6.
+* The `NewOrData` data type has been renamed to `DataFlavor` and extended to
+  support `type data` declarations:
+
+  ```diff
+  -data NewOrData  = NewType | Data
+  +data DataFlavor = NewType | Data | TypeData
+  ```
+
+  Where we upholds the following invariants regarding `TypeData`:
```
goldfirere, line 15:
```suggestion
  We uphold the following invariants regarding `TypeData`:
```

goldfirere: Although this is a bit of a lie, because the type is not abstract. Nothing stops a user from violating this invariant. So it's really a post-condition of the desugaring process, not an invariant, I think.

Language/Haskell/TH/Desugar/AST.hs
```diff
@@ -105,25 +105,47 @@ data DLetDec = DFunD Name [DClause]
...
+            --
+            -- * If the 'DataFlavor' is 'TypeData', then there will be no
+            --   'DDerivClause's, the 'DCxt' will be empty, and the 'DConFields'
+            --   in each 'DCon' will be a 'NormalC' where each 'Bang' is equal
+            --   to @Bang 'NoSourceUnpackedness' 'NoSourceStrictness'@.
+          | DDataD DataFlavor DCxt Name [DTyVarBndrUnit] (Maybe DKind) [DCon] [DDerivClause]
           | DTySynD Name [DTyVarBndrUnit] DType
           | DClassD DCxt Name [DTyVarBndrUnit] [FunDep] [DDec]
-            -- | Note that the @Maybe [DTyVarBndrUnit]@ field is dropped
+            -- | A data family instance declaration. Note the following
```
goldfirere, line 131: This comment looks like it's on the wrong constructor.

goldfirere (approved):
✅

CHANGES.md
RyanGlScott: A fair point. I've changed this to say "Desugaring upholds the following properties regarding `TypeData`" instead. I've adopted similar phrasing in the Haddocks for `DDataD`.

Language/Haskell/TH/Desugar/AST.hs
RyanGlScott: Oops, good catch! I've moved this to `DDataInstD` where it belongs.

PR #180, opened 2023-02-27
Use `DeriveLift` instead of TH to define `Lift` instances

### LINKED ISSUE #176: Use `DeriveLift` instead of TH to define `Lift` instances
Currently, all of the `Lift` instances in `th-desugar` are defined as orphans in a separate [`Language.Haskell.TH.Desugar.Lift`](https://github.com/goldfirere/th-desugar/blob/92c07bd88b74efae0d1b6b7bb5d52049723a009a/Language/Haskell/TH/Desugar/Lift.hs) module. Historically, this made sense, as it also defined orphan instances for external data types in the `template-haskell` library. Nowadays, this is no longer the case, as it simply relies on [`th-orphans`](https://hackage.haskell.org/package/th-orphans) to obtain `Lift` instances for the `template-haskell` library.

Moreover, all of the instances in `Language.Haskell.TH.Desugar.Lift` are defined using Template Haskell by way of the [`th-lift`](https://hackage.haskell.org/package/th-lift) library. Nowadays, this is overkill, as we can just as well use the `DeriveLift` language extension to accomplish the same thing without TH. `DeriveLift` was first introduced in GHC 8.0, which is now the minimum GHC version that `th-desugar` supports.

I propose replacing all of the TH-based instances in `Language.Haskell.TH.Desugar.Lift` with ones generated using `DeriveLift`. There is one complication with this plan, however: the `th-desugar` AST relies on the [`Specificity` compatibility shim](https://github.com/glguy/th-abstraction/blob/3227c489993bd2f5340f69a4ecbeb863be077b81/src/Language/Haskell/TH/Datatype/TyVarBndr.hs#L98-L101) from the [`th-abstraction`](https://hackage.haskell.org/package/th-abstraction) library on pre-9.0 versions of GHC, and that shimmed data type does not define a `Lift` instance. The most direct solution to this issue would be to define a `Lift` instance in the `th-abstraction` library—I've opened glguy/th-abstraction#97 for this.

RyanGlScott:
Now that we depend on GHC 8.0 as the minimum, we can use `DeriveLift` to derive all of our `Lift` instances instead of having to define them as orphan instances using Template Haskell.

This requires `th-abstraction-0.5.*` or later, as that is the first `th-abstraction` version to define a `Lift` instance for its backported `Specificity` data type.

Fixes #176.

### DIFF
--- a/Language/Haskell/TH/Desugar/AST.hs
+++ b/Language/Haskell/TH/Desugar/AST.hs
@@ -6,15 +6,17 @@ Defines the desugared Template Haskell AST. The desugared types and
 constructors are prefixed with a D.
 -}
 
-{-# LANGUAGE CPP, DeriveDataTypeable, DeriveFunctor, DeriveGeneric #-}
+{-# LANGUAGE CPP, DeriveDataTypeable, DeriveFunctor, DeriveGeneric, DeriveLift #-}
 
 module Language.Haskell.TH.Desugar.AST where
 
 import Data.Data hiding (Fixity)
 import GHC.Generics hiding (Fixity)
 import Language.Haskell.TH
+import Language.Haskell.TH.Instances ()
+import Language.Haskell.TH.Syntax (Lift)
 #if __GLASGOW_HASKELL__ < 900
-import Language.Haskell.TH.Datatype.TyVarBndr (Specificity)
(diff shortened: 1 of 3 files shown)


goldfirere (commented):
:tada:

PR #181, opened 2023-03-06
`getDataD`: Also return a `DataFlavor`

RyanGlScott:
This requires moving the definition of `DataFlavor` to `L.H.TH.Desugar.Util` in order to avoid import cycles.

This is ultimately motivated by an eventual fix for goldfirere/singletons#559. While this is a breaking API change, `singletons` is the only user of `getDataD` that I am aware of.


goldfirere (approved):
I trust your judgment about whether this is wise. But please do include a major version bump with this change.

RyanGlScott:
The current Hackage version of `th-desugar` is 1.14, and I've already bumped the major version in the repo to 1.15 in c33d1dfae5e064be1c143929fd65ce9a10412468.

PR #189, opened 2023-06-25
Define `toposortKindVarsOfTvbs`, use it to fix #188

### LINKED ISSUE #188: "Conflicting definitions for ‘a’" error when desugaring data type
The following program:

```hs
{-# LANGUAGE DataKinds #-}
{-# LANGUAGE GHC2021 #-}
{-# LANGUAGE TypeFamilies #-}
{-# LANGUAGE UndecidableInstances #-}
{-# OPTIONS_GHC -ddump-splices -dsuppress-uniques #-}
module Bug where

import Data.Singletons.TH.Options
import Data.Singletons.Base.TH

$(withOptions defaultOptions{genSingKindInsts = False} $
  singletons [d|
  data D a (x :: Maybe a) = MkD
  |])
```

Will cause `singletons-th` to emit ill-scoped code:

```
Bug.hs:(12,2)-(15,5): Splicing declarations
    withOptions defaultOptions {genSingKindInsts = False}
      $ singletons [d| data D a (x :: Maybe a) = MkD |]
  ======>
    data D a (x :: Maybe a) = MkD
    type MkDSym0 :: forall a a (x :: Maybe a). D a (x :: Maybe a)
    type family MkDSym0 :: D a (x :: Maybe a) where
      MkDSym0 = MkD
    data SD :: forall a (x :: Maybe a). D a x -> Type
      where
        SMkD :: forall a a (x :: Maybe a). SD (MkD :: D a (x :: Maybe a))
    type instance Sing @(D a x) = SD
    instance SingI MkD where
      sing = SMkD

Bug.hs:12:2: error:
    Conflicting definitions for ‘a’
    Bound at: Bug.hs:(12,2)-(15,5)
              Bug.hs:(12,2)-(15,5)
   |
12 | $(withOptions defaultOptions{genSingKindInsts = False} $
   |  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...

Bug.hs:12:2: error:
    Conflicting definitions for ‘a’
    Bound at: Bug.hs:(12,2)-(15,5)
              Bug.hs:(12,2)-(15,5)
   |
12 | $(withOptions defaultOptions{genSingKindInsts = False} $
   |  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^...
```

Note the `forall a a` in the type of `SMkD`.

This bug is similar to goldfirere/singletons#567, but it is caused by a different part of the code. (Note that you don't need a standalone kind signature to trigger this bug.)

RyanGlScott:
`toposortKindVarsOfTvbs` ensures that type variables bound earlier in a telescope do not get returned as free variables, thereby fixing the issue observed in #188.

### DIFF
--- a/Language/Haskell/TH/Desugar/Core.hs
+++ b/Language/Haskell/TH/Desugar/Core.hs
@@ -26,6 +26,8 @@ import Control.Monad.Zip
 import Data.Data (Data)
 import Data.Either (lefts)
 import Data.Foldable as F hiding (concat, notElem)
+import Data.Function (on)
+import qualified Data.List as L
 import qualified Data.Map as M
 import Data.Map (Map)
 import Data.Maybe (isJust, mapMaybe)
@@ -959,7 +961,7 @@ dsCon univ_dtvbs data_type con = do
             expl_dtvbs = changeDTVFlags SpecifiedSpec univ_dtvbs ++
                          ex_dtvbs
             impl_dtvbs = changeDTVFlags SpecifiedSpec $
-                         toposortTyVarsOf $ mapMaybe extractTvbKind expl_dtvbs in
+                         toposortKindVarsOfTvbs expl_dtvbs in
         DCon (impl_dtvbs ++ expl_dtvbs) dcxt n fields data_type
       Just gadt_type ->
         let univ_ex_dtvbs = dtvbs in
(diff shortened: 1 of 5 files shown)


Language/Haskell/TH/Desugar/Core.hs
```diff
@@ -1679,6 +1681,24 @@ toposortTyVarsOf tys =
...
+-- this:
+--
+-- @
+-- 'toposortKindVarsOfTvbs' [a :: k, b :: Proxy a]
+-- @
+--
+-- Will return @[k]@, not @[k, a]@, since @a@ is bound earlier by @a :: k@.
+toposortKindVarsOfTvbs :: [DTyVarBndr flag] -> [DTyVarBndrUnit]
+toposortKindVarsOfTvbs tvbs =
+  L.deleteFirstsBy
```
goldfirere, line 1697: This isn't guaranteed to always work. What if you have `forall (a1 :: a2) (a2 :: a1). blah`? This has `a2` free and then binds another `a2` that shadows whatever was already in scope. For names as generated by TH, I think this is probably OK (because those names have uniques), but if there are names generated by `mkName`, I think this won't quite work. I don't think there's a way other than doing some kind of fold.

goldfirere: (A reasonable response is to say that we don't care about the `mkName` case, but then do document the infelicity.)

goldfirere (changes requested):

Language/Haskell/TH/Desugar/Core.hs
RyanGlScott: Good catch. I suppose `toposortKindVarsOfTvbs [(a1 :: a2), (a2 :: a1)]` should just return `[a2]`? If so, I think the following implementation would suffice:

```hs
toposortKindVarsOfTvbs tvbs =
  foldr (\tvb kvs ->
          foldMap (\t -> toposortTyVarsOf [t]) (extractTvbKind tvb) ++
          L.deleteBy ((==) `on` dtvbName) tvb kvs)
        []
        (changeDTVFlags () tvbs)
```

I'll push a version that uses this.

goldfirere (approved):

PR #190, opened 2023-06-29
Handle special tuples in `tupleNameDegree_maybe`, remove `tupleDegree_maybe` and friends

### LINKED ISSUE #187: `tupleDegree_maybe` doesn't take `Unit`, `Solo`, `MkSolo`, or `Tuple` into account
After [GHC!9821](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9821), GHC will desugar `''()` into `''Unit`, `''(,)` into `''Tuple2`, and so on. `th-desugar`'s [`tupleDegree_maybe`](https://github.com/goldfirere/th-desugar/blob/4fabbbeea71f5b2ee1c8e781a3c8bde770efcfcb/Language/Haskell/TH/Desugar/Util.hs#L142-L150) function, on the other hand, does not account for this possibility. As a result, if you try to call `tupleNameDegree_maybe ''()` on GHC HEAD, it will return `Nothing`, since `tupleDegree_maybe` fails to parse `"Unit"`.

I propose that `tupleDegree_maybe` be extended such that:

* `tupleDegree_maybe "Unit"` returns `Just 0`
* `tupleDegree_maybe "Solo"` returns `Just 1`
* `tupleDegree_maybe "MkSolo"` returns `Just 1`
* `tupleDegree_maybe "Tuple<N>"` returns `Just <N>`

Strictly speaking, only the `Unit` and `Tuple<N>` cases are needed to parse TH-quoted tuple syntax on GHC HEAD, since there is no special syntax for `Solo` or `MkSolo`. Still, it would be nice to include `Solo`/`MkSolo` as well for consistency's sake.

RyanGlScott:
This patch:

* Changes `tupleNameDegree_maybe` to handle special tuple names introduced in recent GHC versions, such as `Unit`, `Solo`, `MkSolo`, `Tuple2`, etc. This fixes #187.

* Removes the `tupleDegree_maybe`, `unboxedSumDegree_maybe`, and `unboxedTupleDegree_maybe` functions in favor of the `tupleNameDegree_maybe`, `unboxedSumNameDegree_maybe`, and `unboxedTupleNameDegree_maybe` functions, respectively. This is motivated by the fact that the semantics for `tupleDegree_maybe` have become quite questionable, since it could be possible for a user to supply a custom data type named `Tuple2` that is not the same `Tuple2` as in `GHC.Tuple`. The only way to check for this is to look at the package and module in a `Name`, not a raw string.

  Luckily, removing these functions do not cause much (if anything) to break, as they were only used to power `tupleNameDegree_maybe` and friends.


Language/Haskell/TH/Desugar/Util.hs
```diff
@@ -139,15 +141,39 @@ nameMatches n1@(Name occ1 flav1) n2@(Name occ2 flav2)
...
+-- for the sake of completeness.)
 tupleDegree_maybe :: String -> Maybe Int
-tupleDegree_maybe s = do
-  '(' : s1 <- return s
-  (commas, ")") <- return $ span (== ',') s1
-  let degree
-        | "" <- commas = 0
-        | otherwise    = length commas + 1
-  return degree
+tupleDegree_maybe s
```
goldfirere, line 161: Given that our internal use of this function is solely via `tupleNameDegree_maybe` -- which has more information -- would it make sense to kill off this function in favor of that one? Presumably GHC desugars `'()` into a `Name` that has module information, and so putting this extra login in `tupleNameDegree_maybe` would avoid the possibility of false positives. What do you think?

goldfirere (approved):
Other than a different design possibility, looks good.

Language/Haskell/TH/Desugar/Util.hs
RyanGlScott: > Given that our internal use of this function is solely via `tupleNameDegree_maybe` -- which has more information -- would it make sense to kill off this function in favor of that one?

Good idea. Removing `tupleDegree_maybe` is technically a breaking change, but then again, I can't find _any_ uses of `tupleDegree_maybe` on https://hackage-search.serokell.io (aside from its use in `tupleNameDegree_maybe`), so I don't feel bothered about removing it.

RyanGlScott: I've pushed a new version of this PR that removes `tupleDegree_maybe` (as well as its sibling functions `unboxedSumDegree_maybe` and `unboxedTupleDegree_maybe`).

Language/Haskell/TH/Desugar/Util.hs
```diff
@@ -139,25 +149,80 @@ nameMatches n1@(Name occ1 flav1) n2@(Name occ2 flav2)
...
-        | otherwise    = length commas + 1
-  return degree
-
--- | Extract the degree of a tuple name
+-- | Extract the degree of a tuple 'Name'.
+--
+-- In addition to recognizing tuple syntax (@''(a, b, ..., z)@), this also
+-- recognizes the following:
+--
+-- * @''Unit@ (for 0-tuples)
```
goldfirere, line 157: I see that the tests check for this, but I don't see how the code detects this case. Help?

RyanGlScott: That's because I never actually got around to running the tests with GHC 9.8! You are quite right—there is a missing case for `Unit`.

RyanGlScott: OK, _now_ I've had a chance to test this with GHC 9.8. After adding the missing case for `Unit`, I've confirmed that everything works as expected.

Issue #186, opened 2023-06-09
Adapt to `TypeAbstractions` in GHC 9.8


RyanGlScott:
GHC recently landed [!9480](https://gitlab.haskell.org/ghc/ghc/-/merge_requests/9480), an implementation of [GHC proposal #425](https://github.com/ghc-proposals/ghc-proposals/blob/master/proposals/0425-decl-invis-binders.rst). This introduces invisible `@k`-binders for type-level declarations, guarded behind the `TypeAbstractions` extension. For example:

```
data D @k @j (a :: k) (b :: j) = ...
       ^^ ^^
```

This comes with corresponding changes to the Template Haskell AST, as described in the [GHC 9.8 Migration Guide](https://gitlab.haskell.org/ghc/ghc/-/wikis/migration/9.8?version_id=dd5e567e0dffc04da493e1ba83da408d7f94a2f3#invisible-binders-in-type-level-declarations):

```diff
 data Dec
   = ...
-  | DataD Cxt Name [TyVarBndr ()] (Maybe Kind) [Con] [DerivClause]
-  | NewtypeD Cxt Name [TyVarBndr ()] (Maybe Kind) Con [DerivClause]
-  | TypeDataD Name [TyVarBndr ()] (Maybe Kind) [Con]
-  | TySynD Name [TyVarBndr ()] Type
-  | ClassD Cxt Name [TyVarBndr ()] [FunDep] [Dec]
-  | DataFamilyD Name [TyVarBndr ()] (Maybe Kind)
+  | DataD Cxt Name [TyVarBndr BndrVis] (Maybe Kind) [Con] [DerivClause]
+  | NewtypeD Cxt Name [TyVarBndr BndrVis] (Maybe Kind) Con [DerivClause]
+  | TypeDataD Name [TyVarBndr BndrVis] (Maybe Kind) [Con]
+  | TySynD Name [TyVarBndr BndrVis] Type
+  | ClassD Cxt Name [TyVarBndr BndrVis] [FunDep] [Dec]
+  | DataFamilyD Name [TyVarBndr BndrVis] (Maybe Kind)
   | ...

 data TypeFamilyHead =
-  TypeFamilyHead Name [TyVarBndr ()]      FamilyResultSig (Maybe InjectivityAnn)
+  TypeFamilyHead Name [TyVarBndr BndrVis] FamilyResultSig (Maybe InjectivityAnn)
```

Where `BndrVis` is defined as:

```hs
[...]

PR #194, opened 2023-08-30
Support invisible binders in type-level declarations

RyanGlScott:
Among other changes, there is now a `DTyVarBndrVis` synonym, similar to the existing `DTyVarBndrSpec` and `DTyVarBndrUnit` synonyms, and various types in `Language.Haskell.TH.Desugar.AST` are now defined in terms of `DTyVarBndrVis` instead of `DTyVarBndrUnit`. See the `CHANGES` file for the full details.

This bumps the lower version bounds on `th-abstraction` to `>=0.6`, as this patch makes critical use of several compatibility shims introduced in that version of `th-abstraction`.

Fixes https://github.com/goldfirere/th-desugar/issues/186.

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -8,9 +8,9 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.16.6
+# version: 0.17.20230826
 #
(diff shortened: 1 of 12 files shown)


RyanGlScott:
I'll merge this next week unless there are any review suggestions.

PR #198, opened 2023-09-14
Allow th-abstraction 0.6

Vekhir:
Builds and tests fine


RyanGlScott (approved):
Ah, thanks for the PR! I meant to revise `th-desugar-1.15` on Hackage earlier to allow building with `th-abstraction-0.6`, but I forgot to do so. I've applied this Hackage revision now.

PR #234, opened 2024-12-05
Support ghc 9.12

erikd:


RyanGlScott:
Superseded by 75e15255efab28a4652f359110c6a200525c5aae.

I'll hold off on making a new `th-desugar` release until the corresponding changes in `singletons` are prepared (see https://github.com/goldfirere/singletons/issues/619).

erikd:
@RyanGlScott Wondering if I could have a `template-haskell` upper bound bump on Hackage so I can remove this from the `allow-newer`stanza in my `cabal.project` file?

RyanGlScott:
The current version of `th-desugar` on Hackage doesn't properly support `template-haskell-2.23.0.0`, as that version of `template-haskell` introduces several additions to the AST (`OrP`, `ForallE`, `ForallVisE`, and `ConstrainedE`) that will simply cause `th-desugar` to crash if it encounters them. I'd prefer to make a new, proper Hackage release of `th-desugar` that handles these rather than revising the existing Hackage release into a subtly broken state.

RyanGlScott:
Relatedly, if you'd like to help, I'm trying to figure out what is causing the CI issues [here](https://github.com/goldfirere/singletons/pull/620#issuecomment-2528014758). I think that is the only remaining blocker for releasing new version of `th-desugar`/`singletons`/etc. (which I usually release in tandem) to Hackage.

PR #240, opened 2026-02-06
Use a deterministic counter to generate fresh unique names

mpickering:
The
[`newUniqueName`](https://hackage-content.haskell.org/package/th-desugar-1.19/docs/Language-Haskell-TH-Desugar.html#v:newUniqueName) function has always been rather hacky. Its intended purpose is for generating `Name`s whose `nameBase`s are guaranteed to be unique, even across different splices. It does so in a very brutal fashion:

```
Lines 88 to 94 in [fb1a3b1](/goldfirere/th-desugar/commit/fb1a3b142d15c8c2bc38372da7c0698ec9a6906d)
 -- | Like newName, but even more unique (unique across different splices),
 -- and with unique @nameBase@s. Precondition: the string is a valid Haskell
 -- alphanumeric identifier (could be upper- or lower-case).
 newUniqueName :: Quasi q => String -> q Name
 newUniqueName str = do
   n <- qNewName str
   qNewName $ show n
```

That is, it generates a unique number and creates a `Name` that includes the number as part of its `nameBase`. This means that you will have `Name`s like `a_123456789` in your TH-generated code. It's gross, but it is unfortunately necessary sometimes. See the discussion in [goldfirere/singletons#270
(comment)](https://github.com/goldfirere/singletons/issues/270#issuecomment-342857528) for one example.

While `newUniqueName` does solve the problem of generating fully unique names, it comes will two major drawbacks:

    1. The TH-generated code will become much harder to read due to the unique numeric suffixes. This not only affects `-ddump-splices` output, but it can also affect Haddocks, as is the case with `singletons-th`.

    2. `newUniqueName`'s implementation relies on GHC's unique number generation, which is non-deterministic. As such, the output of `newUniqueName` is also non-deterministic.

When the topic of auditing `newUniqueName` come up last in [goldfirere/singletons#270](https://github.com/goldfirere/singletons/issues/270), only drawback (1) was considered, and the conclusion was that it wasn't worth reimplementing `newUniqueName` just for the sake of generating prettier `Name`s. Drawback (2) is a much more serious one, however. It was first brought up in
[goldfirere/singletons#628](https://github.com/goldfirere/singletons/issues/628), where it was noted that the compilation results of `singletons-th` are non-deterministic. This is pretty dire, so we should strongly consider doing something about this.

This PR implements one solution to this proposal: instead of relying on GHC's unique number generation, we instead generate our own unique numbers to affix as numeric suffixes. This is actually not as difficult to implement as one might expect, as Template Haskell's `Quasi` class lets you store arbitrary state using the `qGetQ`/`qPutQ` methods, and we can store the most recently used unique number using this technique. Whenever one calls `newUniqueName`, the current unique is incremented as then stored back in the state. This solves both drawbacks listed above. It would solve drawback (1), as the unique numbers we would generate would typically be quite small (and therefore easier to read), and it would solve drawback (2), as this technique will always generate the same unique numbers across all compilation runs.

Fixes #239


mpickering:
@RyanGlScott I updated this PR to expose `nextUnique` and `UniqueCounter` so that `singletons-th` can reuse them.

RyanGlScott (approved):
Sorry for the delay on this, @mpickering! This looks great. I'm going to push some minor, mostly documentation-related commits to this and https://github.com/goldfirere/singletons/pull/629 before landing them, but I plan to do the merge shortly thereafter.

